PR #44, opened 2016-08-30
Adam's Work

marxian:


api/public/loo.js
```diff
@@ -34,12 +34,35 @@ routes.loos_in = {
   handler: function * () {
     var loos = yield Loo.findIn(this.params.sw, this.params.ne, this.params.nw, this.params.se).exec()
     this.status = 200
-    this.body = new LooList(loos)
+    this.body = loos
```
marxian: Why not use a LooList here?

models/backup.js
```diff
@@ -0,0 +1,27 @@
+looSchema.methods.updateArea = function * (){
```
marxian: Odd file name? A leftover?

PR #173, opened 2018-06-22
Addition of siphon package for processing loo reports into loos

Hoolean:


packages/siphon/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "version": "2.0.0",
+  "description": "Flexibly and efficiently process loo reports to yield loos",
+  "private": true,
+  "main": "src/index.js",
+  "license": "MIT",
+  "dependencies": {
+    "geolib": "2.0.24",
+    "lodash": "4.17.10",
+    "uuid": "3.2.1"
+  },
```
marxian: We should declare a `devDependency` on `jest`

packages/siphon/src/geo.js
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * This is messy and I don't really like it; hopefully it will refactor out
+ * nicely when this module is tidied.
+ */
+exports.removeBoundWrapAround = function(bounds) {
+  if (bounds.min.lat > bounds.max.lat) {
+    // deep clone
+    const lowBounds = { ...bounds };
+    lowBounds.min = { ...bounds.min };
+    lowBounds.max = { ...bounds.max };
```
marxian:
```
const lowBounds = {
  ...bounds,
  min: {...bounds.min},
  max: {...bounds.max},
};
```

marxian:
Thanks @Hoolean this is well-considered work and puts the project on a good footing for our next stages.

PR #231, opened 2018-07-04
Nearby Loo Improvements

### LINKED ISSUE #185: API near loos endpoint's limit parameter includes inactive loos, which aren't given in the response
## Unexpected Behaviour 

`GET /api/loos/near/0/52?limit=5` gives 4 loos.
`GET /api/loos/near/0/52?limit=100`, with the same default radius, gives 96 loos. This shows that the previous request should have provided 5 loos instead of 4.

## Suspected Cause

The limit parameter is [applied before inactive loos are filtered out](https://github.com/neontribe/gbptm/blob/3a26cc28441c67dc9cf3907808155fe9d9e3f728/packages/api/src/models/loo.js#L10-L26).

## Suggested Remedy

The `$geoNear` query must be first in the aggregation, so a fix is not as simple as changing the order. 

We should consider the function of the `limit` parameter, as it is currently been worked around in our front end by being given an arbitrarily large value instead of being useful.

Perhaps it makes more sense to limit the radius of the search instead of the quantity of results, as this would pertain to how the endpoint is actually used better.

### LINKED ISSUE #186: API incorrectly calculates what loos are in a given radius / bounding box
## Problem

At present, [metres are converted into radians](https://github.com/neontribe/gbptm/blob/3a26cc28441c67dc9cf3907808155fe9d9e3f728/packages/api/src/models/loo.js#L15) before being passed to Mongo's `$geoNear`.

We don't really want to represent our distance as radians while providing an interface that uses metres, however, as the amount of metres a radian represents in longitude varies greatly at different latitudes. Our simple conversion method doesn't take this into account and so the results returned are misleading.

As we use `$geoNear` with a radius to do our bounding box queries too, using [the same method of conversion](https://github.com/neontribe/gbptm/blob/3a26cc28441c67dc9cf3907808155fe9d9e3f728/packages/api/src/models/loo.js#L60), searches within a bounding box of coordinates are also affected.

## Suggested Remedy

Do we want to be using metres for toilet range searches in our API?

* For nearby toilets, almost definitely, as this represents how accessible toilets are for a user far better than coordinates could. [Mongo](https://docs.mongodb.com/manual/reference/operator/aggregation/geoNear/#definition) has support for metre radii, which we should try and use here.
* For coordinate bounds, it is more complicated. We should see how this could be used on the front-end before fixing this implementation but, to fix this, either:
  * The approach for nearby toilets could be used as previously (and clipped to the bounding box, which is _not_ done at the moment, although this is a separate issue) **or**
  * Simple numerical conditions could be tested against the toilet's coordinates (greater than / less than) to eliminate toilets outside the bounds, assuming they are rectangular, which is not checked for at present)

Hoolean:
closes #185

The quantity limit has been removed and so this is no longer an issue. Nearby loos are now only limited by distance, which has been given a maximum and a modest default within the API.

closes #186

Our current API function [deconstructs its limit arguments](https://github.com/neontribe/gbptm/blob/3277b5d6dc95833db70303d94fd15c1c54b09391/packages/ui/src/api.js#L8) but, despite this, everywhere it is called [we are passing a single number instead of an object](https://github.com/neontribe/gbptm/blob/3277b5d6dc95833db70303d94fd15c1c54b09391/packages/ui/src/redux/sagas/loos.js#L24). This was leading us to have both limit and radius as undefined within the function, hitting the API endpoint with an empty query value for limit and hence falling back on what was then the default of 50km.

This behaviour has been made explicit, with the interface also being more clear that if no nearby loos are found, it is because there are none within 50km. This is because, although unintended, this method was working satisfactorily and our code for adjusting the radius relative to our viewport size was a little unsafe in that it was not guaranteed to achieve full coverage with the map projection that most online maps use.

We have a naive approach, as a new query will be issued following every movement, even if the map's viewport has not left the bounds searched in the previous query. I will make a low-priority issue to improve on this, as caching here would lead to significant performance improvements.

### DIFF
--- a/packages/api/src/models/loo.js
+++ b/packages/api/src/models/loo.js
@@ -14,6 +14,7 @@ looSchema.statics.findNear = function(lon, lat, radius) {
         distanceField: 'distance',
         maxDistance: radius,
         spherical: true,
+        limit: 2 ** 62, // infeasibly large number
       },
     },
     {
--- a/packages/api/src/models/loo.js
+++ b/packages/api/src/models/loo.js
@@ -3,7 +3,7 @@ const looSchema = require('./loo_schema').looSchema;
 const _ = require('lodash');
 var Loo;
 
-looSchema.statics.findNear = function(lon, lat, maxDistance, limit) {
+looSchema.statics.findNear = function(lon, lat, radius) {
   return this.aggregate([
     {
       $geoNear: {
@@ -12,8 +12,7 @@ looSchema.statics.findNear = function(lon, lat, maxDistance, limit) {
           coordinates: [lon, lat],
         },
         distanceField: 'distance',
-        maxDistance: maxDistance,
-        limit: limit,
+        maxDistance: radius,
         spherical: true,
       },
     },
--- a/packages/api/src/models/loo.js
+++ b/packages/api/src/models/loo.js
@@ -1,18 +1,19 @@
 const mongoose = require('mongoose');
 const looSchema = require('./loo_schema').looSchema;
 const _ = require('lodash');
-const earth = 6731000;
 var Loo;
 
 looSchema.statics.findNear = function(lon, lat, maxDistance, limit) {
   return this.aggregate([
     {
       $geoNear: {
-        near: [lon, lat],
+        near: {
+          type: 'Point',
+          coordinates: [lon, lat],
+        },
         distanceField: 'distance',
-        maxDistance: maxDistance / earth,
+        maxDistance: maxDistance,
         limit: limit,
-        distanceMultiplier: earth,
         spherical: true,
       },
     },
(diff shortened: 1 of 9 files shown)


packages/api/src/models/loo.js
```diff
@@ -1,19 +1,20 @@
...
+          type: 'Point',
+          coordinates: [lon, lat],
+        },
         distanceField: 'distance',
-        maxDistance: maxDistance / earth,
-        limit: limit,
-        distanceMultiplier: earth,
+        maxDistance: radius,
         spherical: true,
+        limit: 2 ** 62, // infeasibly large number
```
Hoolean: Is there a better workaround here?

marxian: @Hoolean I guess an estimation which inflates the current maximum possible results (c. 12k) by an order of magnitude would also be safe?

Issue #223, opened 2018-07-03
double-spaced caps. 


gailramster:
Writing in caps in 'notes' field returns double-spaced words eg. M O N D A Y S.

Hoolean:
Should be fixed in #245.

PR #245, opened 2018-07-09
Loo prop humanization

### LINKED ISSUE #233: Reduce quantity or scope of near-loo searches in an appropriate place
## Currently, Undesirable

1. We ask the API server for every loo within 50km after every movement on the home page
1. The zoom level allows very wide areas to be looked at and this can lead to slow queries / unresponsive UI

## Hard Problems

* The area visible at a constant zoom level differs radically across different viewport sizes
* The density of loos differs radically between suburban and rural areas
* The ideal quantity or radius for a search for nearby loos differs radically across different use cases, e.g.
  * Looking for your nearest loo
  * Exploring toilet data across the entire UK
  * Looking for all loos within a local area

## Mitigation

We can at least reduce large API hits at larger zoom levels by doing one 50km request and only requesting again when the user moves outside of these bounds or could have traveled enough for there to be a closer loo than their current 5 closest. 

For lesser zoom levels, there is more of an open problem as to how to display useful data while still remaining performant.

Hoolean:
This branch generally overhauls our humanization code for property values, fixing a number of issues:
* [We didn't include properties we didn't know the order of](https://github.com/neontribe/gbptm/commit/f4e1acb9a79de8484ed025d57b38799e8380032b)
* [We humanized all property values, even those that were already human-readable](https://github.com/neontribe/gbptm/commit/6d1dd1d82e5301da0cf429886e3220d600ed39e9), causing issues such as #223 (as a result, this closes #233)

One side-effect of fixing this issue is that the `fee` field may sometimes show as `false`, instead of `No` as before. This is because the `fee` field is, for the most part, human-readable and editable, and there is no practical way to distinguish between a user-entered `false` string and a pre-existing one.

To avoid writing code to accommodate a specific quirk of our data and, at the same time, risk translating user input erroneously, this has been left as-is. If it is considered to be preferable to retain the previous translation, this specific behaviour can be re-added. Worth considering in this is that about 3 in 26 of our loos will show their fee as `false`.

### DIFF
--- a/packages/ui/src/config.js
+++ b/packages/ui/src/config.js
@@ -17,50 +17,53 @@ export default {
   allowAddEditLoo: true,
   showBackButtons: false,
   looProps: {
-    opening: definitions,
-    type: [
-      {
-        name: 'Female',
-        value: 'female',
-      },
-      {
-        name: 'Male',
-        value: 'male',
-      },
-      {
-        name: 'Female and Male',
-        value: 'female and male',
-      },
-      {
-        name: 'Unisex',
-        value: 'unisex',
-      },
-      {
-        name: 'Male Urinal',
-        value: 'male urinal',
-      },
-      {
-        name: 'Children Only',
-        value: 'children only',
-      },
-      {
-        name: 'None',
-        value: 'none',
-      },
-    ],
-    access: [
-      {
-        name: 'Public',
-        value: 'public',
-      },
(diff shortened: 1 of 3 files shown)


Hoolean:
After some discussion, the edge-case discussed above has been written around instead. This should only be in effect until we normalise our database better.

PR #276, opened 2018-07-24
Clustering improvements

### LINKED ISSUE #241: Improvements to Clustering
Currently LooMap takes a prop to enable clustering or not.

This is probably redundant now we use "NearbyLooMap" almost everywhere.

I suspect we should use only the markercluster in LooMap which will make it cleaner.
We should also investigate adding the markers to the markercluster in chunks to gain some better performance.

Please take a look at the markercluster docs here: https://github.com/Leaflet/Leaflet.markercluster
With especial regard to: https://github.com/Leaflet/Leaflet.markercluster#bulk-adding-and-removing-markers

There is also a react wrapper around markercluster, but users are reporting memory leaks - so we'll hold off on using that...

Hoolean:
closes #241

## Cleanup

Clustering was [enabled everywhere](https://github.com/neontribe/gbptm/commit/fa6d54d5d3198a35dd6c0213b86eba72a9403458). Prior to this, we had separate branched implementations for clustered and unclustered `LooMap`s. We no longer had the explicit need to disable clustering, so we were able to scoop up some cheap maintainability by removing the unclustered implementation.

## Performance

Previously, on every render we regenerated all markers. There were a few issues with this:

* Obvious performance hit, especially in areas such as London with over 1000 loos within 50km
* Clustering was not deterministic and, as such, on moving cluster sizes would visibly change even when the loos in the area had not

One would hope that React could be used to reduce some of this effort but, sadly, the `react-leaflet-markercluster` library has [memory leaks reported](https://github.com/YUzhva/react-leaflet-markercluster/issues/70) (which is odd considering [the very small size of the project](https://github.com/YUzhva/react-leaflet-markercluster/blob/master/src/react-leaflet-markercluster.js); all of the garbage collection is handled upstream in `react-leaflet`).

We wanted to update markers instead of recreate them but as `leaflet-markercluster` is not built around React and has a rather state-y nature we must use `componentDidUpdate` to do this as opposed to `render`. [Markers are now being created and destroyed there](https://github.com/neontribe/gbptm/commit/b2ac95fc7479077d81e697e702f6226426bffb59). Caching markers in the component's state and diffing previous and present `loos` prop values massively reduces the number of recreated markers and stops us from having to clear and re-add markers on each render.

## Future Performance

Adding markers in chunks is yet to be tried, although now using one call to `addLayers` and `removeLayers` in each update does mean that it would be as simple as changing `leaflet-markercluster`'s options to enable this.

It is worth considering that chunking would leave us with a blind spot in our UI. At present, the user can clearly see when all loos have been loaded; there are none and then, after a small delay, they are all there. Enabling chunking could potentially show some markers before the rest are added, which would be misleading unless we added UI to show that there were still loos to come.

I am yet to implement chunking because of this potential issue and would be interested to see how the current changes affect performance before we consider implementing it.

### DIFF
--- a/packages/ui/src/components/LooMap.js
+++ b/packages/ui/src/components/LooMap.js
@@ -134,32 +134,30 @@ export class LooMap extends Component {
       var map = (this.leafletElement = this.refs.map.leafletElement);
       var center = map.getCenter();
 
-      if (this.props.shouldCluster) {
-        // Create a marker cluster layer which will be reset each time this
-        // component receives props
-        var clusterLayer = L.markerClusterGroup({
-          showCoverageOnHover: false,
-          disableClusteringAtZoom: 15,
-          iconCreateFunction: cluster => {
-            var count = cluster.getChildCount();
+      // Create a marker cluster layer which will be reset each time this
+      // component receives props
+      var clusterLayer = L.markerClusterGroup({
+        showCoverageOnHover: false,
+        disableClusteringAtZoom: 15,
+        iconCreateFunction: cluster => {
+          var count = cluster.getChildCount();
 
-            return L.divIcon({
-              html: `<div class=${styles.cluster}>
-                                <span class=${styles.count}>${count}</span>
-                                toilets
-                            </div>`,
-            });
-          },
-        });
+          return L.divIcon({
+            html: `<div class=${styles.cluster}>
+                              <span class=${styles.count}>${count}</span>
+                              toilets
+                          </div>`,
+          });
+        },
+      });
 
-        this.setState({
-          clusterLayer,
-        });
-      }
-
-      this.props.onInitialised(map);
-      this.props.onUpdateCenter(center);
+      this.setState({
+        clusterLayer,
+      });
     }
+
+    this.props.onInitialised(map);
+    this.props.onUpdateCenter(center);
   }
 
   onMove(event) {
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/LooMap.js
```diff
@@ -182,6 +210,114 @@ export class LooMap extends Component {
...
+    let loosThen = _.keyBy(prevProps.loos, '_id');
+    let loosNow = _.keyBy(this.props.loos, '_id');
+
+    // Remove elements that have unchanged location from both, effectively
+    // diff'ing; this will leave us with markers to remove in loosThen and
+    // markers to add in loosNow
+    for (let id of Object.keys(loosNow)) {
+      if (
+        loosThen.hasOwnProperty(id) &&
+        _.isEqual(loosThen[id].geometry, loosNow[id].geometry)
```
marxian: We might gain a shred of simplicity/performance here by comparing the geohashes? But wait... Isn't the _id enough?

Hoolean: There are two potential causes of the `loos` prop updating:

1. The set of nearby loos has changed (e.g. if the user has moved) _and/or_
1. A loo's properties have been changed (e.g. after the user has just edited it)

Checking just `_id` is enough to ensure that we keep the view up-to-date for the first case but we must compare location to keep the view up-to-date for the second.

While it's not ideal to remove and re-add a loo to handle it moving, this should rarely be done more than once per call to the function so it should be fine for now.

On the other hand, we are checking equality of the geometry object for as many loos as there are within 50km on each function call, so it's probably worth moving to `geohash` there; I'll get a commit up for that.

PR #287, opened 2018-07-26
feat: scope panning and zooming abilities in a friendly & consistent way

### LINKED ISSUE #266: Cannot pan map in toilet report view
When viewing a toilet report, other toilet pins now show (helpful) but you cannot pan the map. This was a useful feature to identify where the toilet is in relation to you, when trying to find it, and to find the next nearest toilet to you (from Phillip's feedback). 
Can it be reactivated or is there a design-reason for locking the map?

Hoolean:
closes #266

We now let the user pan and zoom to their heart's content everywhere, with the exceptions of:

* **Add/edit page** - minimum zoom is set; we expect a certain level of precision when specifying locations
* **Remove page** - panning and zooming is disallowed, minimum zoom is set; we expect the user to stay focused on the single loo they are removing

The ability to pan and zoom around does introduce a few scenarios in which the location of the selected loo becomes unclear. If the loo leaves the viewport or is clustered when the user zooms out (the minimum zoom now allows clustering while a loo is selected) then there is no visual cue to indicate where the selected loo is. I would be interested to hear thoughts on this before we merge and mark the issue as resolved.


marxian:
Maybe we should expose the `disableClusteringAtZoom` property of the clusterlayer as a prop on the LooMap, so we can use it to stop clustering in this view?

PR #288, opened 2018-07-27
fix: show newlines in loo property values

### LINKED ISSUE #253: Newlines are ignored when showing Notes
## Problem

Loos can have Notes, as given in a textarea, and these Notes may be written across several lines. Unfortunately, when we display these Notes with the toilet information, all the text is shown on one line.

[e.g.](https://gbptm-unity.herokuapp.com/loos/54c2348c2ec4abe957b83e79)
All on one line:
nl1
...but there are clearly multiple lines in the notes:
nl2

**NOTE: The screen shot shows spaces between capitals. This is from a past issue (#223) that has since been resolved.**

## Additional Notes

While untested, this may also be an issue with the Removal Reason section.

Hoolean:
closes #253

Sometimes property values have newlines in them; this alteration to the CSS ensures that they're displayed.

### DIFF
--- a/packages/ui/src/pages/css/loo-page.module.css
+++ b/packages/ui/src/pages/css/loo-page.module.css
@@ -35,6 +35,10 @@
 	line-height: 1;
 }
 
+.propertyValue {
+  white-space: pre-line;
+}
+
 .dataTable {
 	table-layout: fixed;
 	width: 100%;


packages/ui/src/pages/css/loo-page.module.css
```diff
@@ -35,6 +35,10 @@
 	line-height: 1;
 }
 
+.propertyValue {
+  white-space: pre-line;
```
marxian: I had no idea this was a thing :blush:

Issue #347, opened 2018-08-29
Recently Updated


marxian:
**Is your feature request related to a problem? Please describe.**
We'd like to be able to search for recently updated loos so that we can assess the quality of new reports

**Describe the solution you'd like**
Maybe add a facility to the search page? Or implement a dedicated route?

PR #355, opened 2018-08-31
feat: order by updatedAt, filter by updatedAt

ob6160:
resolves: #347 

* Allows the user to search for toilets updated between a certain date range and order ascending or descending.
* Adds three new form elements with matching params to the search:
  * Search order (ascending, descending) => `order` param, defaults to `desc` (optional)
  * From Date (date picker) => `from_date` param (optional)
  * To Date (date picker) => `to_date` param (optional)

Also includes a small bit of error handling for searches so that the search button is not permanently greyed out if the request falls through.

### DIFF
--- a/packages/admin/src/components/Search.js
+++ b/packages/admin/src/components/Search.js
@@ -44,6 +44,7 @@ class Search extends Component {
 
     var defaults = {
       text: '',
+      area_name: '',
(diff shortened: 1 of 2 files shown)


ob6160:
Closing to make a few changes

PR #369, opened 2018-09-17
Addition of `loodb` package

### LINKED ISSUE #337: Pull report creation code into its own interface
The migration code is getting a little cumbersome, as it's mixing logic for the creation new-schema models and the extraction of old-schema models. Pulling out this new-schema creation logic would make the migration code cleaner and more reusable.

Hoolean:
A `loodb` package has been added; definitions and logic related to the new Mongoose schemas are situated here.

Some methods have been added that, as well as being immediately useful for migration code, should be useful in the future. One can imagine that more will be added when `api` is refactored to use the new schemas.

Closes #337.


packages/migrate/bin/summarise-report-values.js
```diff
@@ -2,7 +2,7 @@ const fs = require('fs');
 
 // the properties that can be in a new report
 const properties = Object.keys(
-  require('../lib/models').Report.schema.tree.diff
+  require('../src/models').Report.schema.tree.diff
```
ob6160: This line causes an error for me - the file seems to have been moved to the `@neontribe/gbptm-loodb` package.

packages/ui/src/pages/AboutPage.js
```diff
@@ -127,7 +127,8 @@ class AboutPage extends Component {
               </p>
```
ob6160: Was this caused by the linter/formater?

packages/migrate/bin/summarise-report-values.js
Hoolean: Correct; I'll point it in the right place else find a better way to source the property keys.

packages/ui/src/pages/AboutPage.js
Hoolean: Yep, didn't intentionally touch anything in `ui`.

PR #375, opened 2018-09-19
371 migration cleanup

### LINKED ISSUE #371: Migration script doesn't cleanup database connections and hangs
A `mongoose` connection is established [within the migration code](https://github.com/neontribe/gbptm/blob/loodb-package/packages/migrate/src/migrate.js#L2-L4) that is never closed. Unfortunately, any of our scripts requiring a file that opens such a connection will never close it and so hang.

We [expose the close function](https://github.com/neontribe/gbptm/blob/loodb-package/packages/loodb/src/index.js#L12) of the connection in the `loodb` package but it is not immediately accessible within the CLI migration scripts; we may need to create our `loodb` connection in these and inject it into the other scripts through a slightly more complicated method to be able to close it when we're done.

ob6160:
resolves #371 
Exposes db connection close method so the migration no longer hangs.


Hoolean:
Have noticed that `summarise-report-values` is also affected; is it possible to easily extend the fix to that too?

PR #382, opened 2018-10-02
feat: edit button on loo view with link to loo edit

edpe:
I'm fairly certain this works. My machine wasn't able to run both admin and ui at the same time so hard to test but copying and pasting the generated link seemed to work ok.


edpe:
also just realised my linter seems to have been rather over-enthusiastic - maybe I need to change settings to match yours?

marxian:
Hmm... I wonder what's gone wrong there. We can take a look at settings tomorrow?

edpe:
Yes please - whenever you have time.

PR #389, opened 2019-01-25
Explorer index page

edpe:
Basic intro page for the explorer

### DIFF
--- a/packages/explorer/src/components/Home.js
+++ b/packages/explorer/src/components/Home.js
@@ -1,11 +1,71 @@
 import React, { Component } from 'react';
+import Paper from '@material-ui/core/Paper';
+import ListItemText from '@material-ui/core/ListItemText';
+import styles from '../css/home.module.css';
 
 class Home extends Component {
   render() {
     return (
-      <div>
-        <h1>Welcome</h1>
-        <p>There will be highlights here one day.</p>
+      <div className={styles.wrapper}>
+        <div className={styles.intro}>
+          <p>
+            Toilet Map Explorer exists to help people who want to know more
+            about the data behind the{' '}
+            <a
+              href="https://www.toiletmap.org.uk"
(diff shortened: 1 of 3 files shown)


packages/explorer/src/components/Home.js
```diff
@@ -1,11 +1,128 @@
...
 
 class Home extends Component {
   render() {
     return (
-      <div>
-        <h1>Welcome</h1>
-        <p>There will be highlights here one day.</p>
+      <div className={styles.main}>
+        <div>
+          <br />
```
falconx: This `br`element is not appropriate for use layout concerns.

packages/explorer/src/components/Home.js
```diff
@@ -1,11 +1,128 @@
...
+  root: {
+    padding: '1rem',
+  },
+})(Typography);
+
+const SubTypography = withStyles({
+  root: {
+    paddingLeft: '1rem',
+    paddinRight: '1rem',
+    paddingBottom: '1rem',
```
falconx: I believe presentation and layout concerns should be separated in CSS.

It seems like the entire content area could do with a wrapping element which applies the lateral gutters and then to apply a `margin` below each each text. I'm not sure if there are other MUI layout components or more global configurations we can set to define these rules?

I also see that, in addition to the spacing at the bottom of each `Typography` that we're also applying a `br` for additional spacing. We should avoid the use of `br` here.

packages/explorer/src/css/home.module.css
```diff
@@ -0,0 +1,9 @@
+body {
+  font-family: Cabin, sans-serif;
+  background-color: #e8eaf6;
+}
```
falconx: We'll also want a CSS reset if we're to ensure it looks the same across all browsers and devices. Though perhaps MUI already implements this for us?

PR #411, opened 2019-07-19
Resolve #407: switch explorer statistics to GraphQL interface

### LINKED ISSUE #407: Port Statistics to GraphQL
Displays such as https://www.toiletmap.org.uk/explorer/statistics currently depend on the rest endpoint defined here: https://github.com/neontribe/gbptm/blob/develop/packages/api/src/rest/statistics.js

We'd like to model this data in the schema and make it available through GraphQL.

I suggest that at the same time we try to tidy up the construction of the mongoose queries used here, and house them in methods on the models where they belong: https://github.com/neontribe/gbptm/blob/develop/packages/api/src/db/loo.js

jthistle:
This aims to eventually address #407. 

I'm opening this PR to allow code review and to share what I'm working on.


packages/api/src/graphql/typeDefs/index.js
```diff
@@ -115,6 +115,18 @@ const typeDefs = gql`
     loo: Loo
   }
 
+  """
+  A container type for various statistical counters
+  """
+  type Counters {
```
marxian: It'b be totally amazing to document these properties... If you can figure from the logic what they _really_ mean ;-)

jthistle:
Right, I think this is pretty much done. All the numbers come through GraphQL now, and any mongo queries are attached to the mongoose schema.

jthistle:
Search is now moved (apart from the queries for the list of areas, which is something else I could now GraphQL-ify I suppose).

packages/api/src/db/report.js
```diff
@@ -38,7 +38,7 @@ const ReportSchema = new Schema(
         }
 
         // check that it references us as the previous report
-        if (!next.previous.equals(this._id)) {
+        if (!next.previous.equaReportSchemals(this._id)) {
```
jthistle: note to self: revert this line

packages/api/src/graphql/resolvers/index.js
```diff
@@ -27,19 +28,60 @@ const looInfoResolver = property => {
 const resolvers = {
   Query: {
     loo: (parent, args) => Loo.findById(args.id),
-    loos: async (parent, args) => {
+    loos: async (parent, args, context) => {
+      console.log('search: ', args);
```
jthistle: note to self: remove this console log

jthistle:
Aha, thank you very much.

PR #413, opened 2019-07-24
feat: add 404 page

### LINKED ISSUE #300: 404 pages
## Problem

Navigating to an unknown route will either take us to:

* A black screen [when no route is matched](https://www.toiletmap.org.uk/qwerty)
* An invalid state [when an invalid parameter is given](http://localhost:3000/loos/qwerty)

We should handle this properly.

jthistle:
Fixes #300.

This adds a 404 page that looks a bit like this:

image

or on mobile:

image

Changes:

- Create a new page, `404`
- Move `Logo` to its own component with its own stylesheet

To get a different background colour on the 404 page, I've used a slightly hacky method of a fixed position div with `100vw` width and `100vh` height and a z-index of `-10`, since I can't change the body background colour.


jthistle:
Conflicts fixed @marxian

jthistle:
FWIW, #414 is also still around and ready for merge

Issue #415, opened 2019-08-12
Tracking


RpprRoger:
## Is your feature request related to a problem? Please describe.
Adding multi level tracking and cookie banner

## Describe the solution you'd like
- Add tracking required for Adobe and GTM suites. The Cookie banner should offer 3 options, our tracking, all tracking, and no tracking.
- The stage the app is running on will decide what sort of tracking script we will load.
- Add some copy to the privacy policy and the popup to describe the actions
- Let the user know that we want to track and let them choose

### The Cookie banner

We currently use localstorage (which counts as "cookies") and cookies for the following reasons:
- We use local storage for core login functionality
- We use local storage to store preferences
- We use cookies via google to track user's devices/locations/pages
- (PLAN) We want to track page views which we will share with Unilever
- (PLAN) To track link clicks which will be shared with unilever

### Levels of tracking:

To me this indicates that there is 3 levels of tracking
1. No Tracking
1. Toilet Map tracking (Google Tag manager)
1. Unilever page tracking & action tracking (Adobe)

## Additional context
There is [some documentation which was provided by Unilever (requires neontribe email) https://drive.google.com/drive/folders/1PPJMOilDLNNJnjSm9qSCELh-FNQdMfFM?usp=sharing](https://drive.google.com/drive/folders/1PPJMOilDLNNJnjSm9qSCELh-FNQdMfFM?usp=sharing)

RpprRoger:
We need some words to talk about the type of tracking we want to do, and a change to the privacy policy. We can see the current privacy policy here: https://www.toiletmap.org.uk/privacy

RpprRoger:
The bulk of this is completed, see #443 for outstanding

PR #416, opened 2019-08-12
feat(#415): add initial tracking scripts and popup

RpprRoger:
Initial setup for tracking and cookie popup, needs copy and more thinking about how toggling the tracking works. (basically just fix the TODOs)

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,1 +1,3 @@
 #--install.pure-lockfile true
+# Always install exact package versions
+--add.exact true
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -29,7 +29,7 @@
     "react": "16.7.0",
     "react-addons-update": "15.6.2",
     "react-dom": "16.7.0",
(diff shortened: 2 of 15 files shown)


packages/ui/package.json
```diff
@@ -29,6 +29,7 @@
     "react": "16.7.0",
     "react-addons-update": "15.6.2",
     "react-dom": "16.7.0",
+    "react-helmet": "^5.2.1",
```
marxian: Pin dependency versions please!

RpprRoger: Thanks, I'll do that now. We could use [that trick](https://github.com/neontribe/www/blob/master/.yarnrc#L7-L8) to make sure that `yarn add`s are exact

.yarnrc
```diff
@@ -1 +1,3 @@
 #--install.pure-lockfile true
+# Always install exact package versions
+--add.exact true
```
marxian: I'd forgotten we could do that!

PR #420, opened 2019-08-14
fix: adding missing adobe javascript

RpprRoger:
The Adobe documentation says to add this extra JS, from what I can tell it just hides the whole web page for 3 seconds? It's pretty gross, I don't recommend we merge this unless absolutely necessary.


jthistle:
Ugh, why would that ever be necessary?

RpprRoger:
> Ugh, why would that ever be necessary?

The _only_ thing I can think of is that a script loaded later expects there to be '#at-body-style' in the DOM.

And that's tenuous

jthistle:
It just seems like a nice way to make sure half your potential users leave the site...

PR #427, opened 2019-10-10
Update README.md to change git clone

charlesstrange2:
Currently, if git isn't setup for ssh you'll get 
```
Cloning into 'gbptm'...
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```
As the repo is public, we don't need access via ssh so we can fetch over https.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ This documentation is oriented towards developers, if you'd like to learn more a
 ### Installation
 
 ```
-git clone git@github.com:neontribe/gbptm.git gbptm
+git clone https://github.com/neontribe/gbptm.git gbptm
 cd gbptm
 nvm install && nvm use
 npm install -g yarn


jthistle:
Maybe a bit late for this, but do we really need to tell people to clone the project at all? Imo, it seems like a pretty obvious prerequisite...

charlesstrange2:
Ha! Yes, this is, of course true ;-)


It didn't take those of us without ssh keys setup correctly long to work out why that message appeared - in my case my overly convoluted IDE had its brain in another project. If I'd just been told "clone it somewhere, use nvm" rather than follow the detailed instructions I might never have had the issue.


However, It's more about perception of professional care.


- the project is open, anyone can play with it

- there are internal and external training uses for it

- getting from zero to something in a couple of hours is a validating experience

- not everyone has the same technical cabilities or experience

- not everyone is using the same tools or methods.


So lowering the barriers to entry a bit and avoiding "gotcha" hurdles for a new contributer without going too far out of our way to do so is polite, friendly leg-up that might endear the company to them.

________________________________

jthistle:
True, I guess. Lowering barriers is good, I agree, up to a point. But at some point, surely things become obvious enough that you don't need to mention them.

At the risk of sounding facetious, don't we need to tell potential contributors how to get git working on their computer, or how to install `npm` and/or `nvm` as well? They're both prerequisites. Do we need to tell them how to connect to the internet? What about how to turn their computer on? ;)

In all seriousness, I see no real reason not to have good instructions in this case, even for something as basic as cloning, but doc-writing would take _forever_ if everything had to be as thorough as that. There's a balance I suppose, as with most things.

Issue #443, opened 2019-10-24
Dynamic tracking data for Adobe Analytics


RpprRoger:
**Describe the bug**
When sending tracking data to adobe analytics we need to add some dynamic data.

Please see the comments in this file https://github.com/neontribe/gbptm/blob/develop/packages/ui/public/adobe-digital-data.js and [line 30](https://github.com/neontribe/gbptm/blob/develop/packages/ui/public/adobe-digital-data.js#L30) for the missing data.

**To Reproduce**
Change the [get stage function](https://github.com/neontribe/gbptm/blob/develop/packages/ui/src/config.js#L24-L34) to return staging, so you can test locally.
Load the toilet map and allow adobe analytics tracking.
Inspect `window.digitalData`.
Dynamic data should be replaced with the data from the browser.
e.g.:
- the `destinationURL` is `window.location.href`
- the `pageName` is `document.title`

**Expected behavior**
The values should be properly populated from the browser

**Additional context**
A good way to detect when a page changes is to use [`history.listen` (docs)](https://github.com/ReactTraining/history/blob/3f69f9e07b0a739419704cffc3b3563133281548/docs/GettingStarted.md#listening)

More context #415

PR #444, opened 2019-10-30
443 dynamic tracking for adobe

charlesstrange2:
As per https://github.com/neontribe/gbptm/issues/443

found a spot to try setting a listener on the history.
Pros:
- does notice a page change
- does fill the `window.digtalData` variable if it's there

Cons
- won't trigger on page _load_
- can't actually tell if it's being sent to adobe. :man_shrugging: can someone check that?


marxian:
@charlesstrange2 I _think_ that the adobe scripts fire off a track when they load anyway...
It seems to me that this'll set the relevant data but won't actually fire a tracking request?

I usually look to see them going out in the network tab...

charlesstrange2:
Will close and re-check to see how we definitely fire off a transmission

PR #463, opened 2019-12-22
GraphQL: the bumper Christmas PR

jthistle:
It works! It's functional! It doesn't crash (well, most of the time)!

In theory I think this is in a mergeable state, but obviously, some testing is much needed. Well, that is, if you can find loos to test with that aren't in the middle of the ocean.

The work is by no means finished, however...

Merry Christmas! :santa:


marxian:
Bravo @jthistle !

I'll check this out in the next couple of days. Happy Christmas!

packages/ui/src/components/WithApolloClient.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
marxian: I think this functionality would be best coming from `@apollo/react-hoc` - unless I've missed the point somewhere?

jthistle:
I didn't realise there was a way to do it without read/writeQuery... do you have a docs link for that?

As for what's left to be done: dereduxing, cleaning up code, testing, and deprecating REST, basically.

marxian:
@jthistle Here's the docs for local schema:

https://www.apollographql.com/docs/react/v3.0-beta/data/local-state/

jthistle:
Thanks! One more thing to do that I forgot to mention is to keep up to date with Apollo and the release of @apollo/client out of beta, which should be soonish.

jthistle:
Ok, after reading the docs it seems I've been doing local state things a bit wrong. I'll rework that bit and tack it onto the end of this PR.

marxian:
Do you need any help with that? It took a while for it to click into place for me...

jthistle:
It's possible... I'll see what I can do without bothering you too much, but if I get stuck I'll let you know :)

jthistle:
@marxian, right I think I've got the local resolvers/schema working now. Just some changes to remove `readQuery` and `writeQuery`. Hope it looks OK!

jthistle:
Now it works with the latest apollo client beta. However, the zoom is still broken and I'm still stuck as to why. @marxian or maybe even @RpprRoger - could one of you check out this branch and have a look at this problem when you have a moment?

jthistle:
Don't worry, just managed to fix it :)

I know 13 lines doesn't look like much, but it took a hell of a lot of debugging to work out what I could do to fix it.

PR #482, opened 2020-03-01
The Big GraphQL + Mobile support launch

marxian:
On merge we must update the production database as follows:

- [ ] `db.getCollection('newloos').update({"properties.access": "customers"}, {$set: {"properties.access": "customers only"}}, {"multi": true})`


jthistle:
Ahhhh! Sorry. I didn't mean to click the big button that said 'mark as ready for review'. Is there any way to set it back to being a draft?

Also, you may want to update `@apollo/client` to `beta.38`. Not that there's any significant changes, just that it may be more stable.

PR #506, opened 2020-03-15
fix: explorer search broken in many ways

### LINKED ISSUE #503: Explorer search is broken
**Describe the bug**
Many search features in the explorer are broken. Changing the dates in advanced settings and pressing search does nothing. Limiting the number of responses doesn't actually change anything. Areas and text do work, however.

This is on develop.

jthistle:
- pagination settings now have an effect
- to make a new query the search button must be pressed, rather than just having
  a half-working automatic query system, which could also cause problems
- fixed an error with setting the page to 1 when there was no data

resolves #503

### DIFF
--- a/packages/explorer/src/components/Search/index.js
+++ b/packages/explorer/src/components/Search/index.js
@@ -114,6 +114,16 @@ const Styled = createStyled(theme => ({
 
 const renderTableRows = ({ data }) => {
   const MISSING_MESSAGE = 'Not Recorded';
+
+  if (data.loos.length === 0) {
+    return <TableRow>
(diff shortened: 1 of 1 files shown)


packages/explorer/src/components/Search/index.js
```diff
@@ -356,8 +370,8 @@ class Search extends Component {
   /**
    * Getter for the search query string - strips empty fields.
    */
-  get queryString() {
-    const omitEmpty = _.pickBy(this.state.searchParams);
+  queryString(newParams) {
```
marxian: :+1:

PR #533, opened 2020-03-27
chore: replace class components with functions in ui

stphnnnn:
Some components are more trivial to convert to functions than others, I would appreciate someone who knows the desired behaviour of the application well to check I haven't broken anything.

I've left the `LooMap` component for now as it's quite complex. I plan to have a function component replacement for this in a future PR.


marxian:
@stphnnnn 

I've had a look around the preview deployment and all seems well, except when I try to add a Loo...

stphnnnn:
There is a bug in this branch around google anlytics. With every route change, a new `analytics.js` is added to the DOM causing pages to get tracked multiple times.

I think this is due to:
- the `Analytics` component being a child of each route rather than a persistent fixture of the application (it does seem that that's what we should be doing, looking at the example: https://github.com/fknussel/react-router-ga#usage-example)
- `react-router-ga` not checking to see if GA has already been initialized before initializing and not doing any cleanup on unmount

I'm not sure why this wasn't happening in the class component version of the `Tracking` which we had previously.

PR #550, opened 2020-04-02
build: flatten monorepo

marxian:


stphnnnn:
this looks good to me, files in /public are being served correctly too.
next step is to nest `explorer` inside th react application right?

marxian:
@stphnnnn that's certainly needed, there are a couple of other issues remaining too.

You can see the state of things in the project: https://github.com/neontribe/gbptm/projects/7

PR #581, opened 2020-04-24
refactor: simplify state flow

stphnnnn:
This PR aims to simplify the state across the application by aiming for a single source of truth for both map and loos state and a clearer flow of state down through components. This is achieved by the following:

1. A single LooMap component which simply renders markers on a map with map controls 
 - This component has been refactored to make better use of JSX syntax, for example the markers are now components which are rendered as a child of the Leaflet Map component
 - Some of the map elements have been split into their own component files to help make it clearer what the LooMap component itself does

2. Some custom hooks which handle reusable business logic throughout the application:
 - useGeolocation: gets the geolocation of the users device 
 - useMapPosition: helper to get and set the global map position state (lng, lat, zoom)
 - useNearbyLoos: accepts map position arguments and uses an Apollo query to get a list of nearby loos

3. Splits the AddEditPage into 2 individual pages to make the behviour of each clearer. This could be improved further by abstracting the form into a reusable component. 

Other changes:
- Simplified the `latLngToDistance` function by using a helper built into Leaflet
- Improved the way Apollo caches and queries for loos which allows for a loo to be displayed on the LooPage map before the extra data has loaded


stphnnnn:
this has been merged into `revenge` branch so is part of #550

PR #584, opened 2020-04-27
refactor: avoid prop drilling with auth context

falconx:


src/index.js
```diff
@@ -91,113 +78,126 @@ const cache = new InMemoryCache({
...
-        mapCenter {
-          lat
-          lng
-        }
-        userData {
-          loggedIn
-          name
-        }
+ReactDOM.render(
+  <AuthProvider>
```
marxian, line 82: I'm confused about doing the apolloClient setup inside the dom renderer.

I'm struggling to put my finger on quite why but it feels... odd somehow?

Am I worrying over nothing?

falconx: I know what you mean. It feels weird to me, so much so I messaged Steve about the render prop approach. Do you know of another solution?

I guess we could hide the setup with something like this:

`index.js`:
```
<AuthProvider>
  <MyApolloProvider>
```

`MyApolloProvider.js`:
```
const auth = useAuth();
const client = new ApolloClient({ .. });

<ApolloClient client={client}>
```

falconx: Perhaps like this https://github.com/neontribe/gbptm/compare/auth...auth-app?expand=1

marxian: Well that's certainly less disturbing!

PR #585, opened 2020-04-27
feat: introduce PageLoading component

falconx:
Not sure if we have designs coming for this?


marxian:
This was more a technical necessity at the moment. We'll need to tell Rachel that this state exists at some point...

PR #588, opened 2020-04-29
refactor: manage EntryForm with react-hook-form

falconx:
Difficult to know, without tests if, whether this actually breaks everything.

* Added `<form>` element
* Abstracted `EntryForm` into shared component for `AddPage` and `EditPage`
* Added `react-hook-form` dependency

### DIFF
--- a/src/components/EntryForm.js
+++ b/src/components/EntryForm.js
@@ -1,6 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import MediaQuery from 'react-responsive';
+import { useForm } from 'react-hook-form';
+import isFunction from 'lodash/isFunction';
 
 import layout from '../components/css/layout.module.css';
 import styles from '../pages/css/edit-loo-page.module.css';
@@ -9,6 +11,7 @@ import headings from '../css/headings.module.css';
 import controls from '../css/controls.module.css';
 
 import config from '../config';
+import history from '../history';
 
 import Loading from '../components/Loading';
 import DismissableBox from './DismissableBox';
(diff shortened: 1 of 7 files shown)


src/pages/AddPage.js
```diff
@@ -88,37 +60,6 @@ const AddPage = (props) => {
...
-  const [looState, setLooState] = useState(getInitialLooState());
-
-  const handleChange = (event) => {
-    setLooState({
-      ...looState,
-      [event.target.name]: event.target.value,
-    });
-  };
-
-  const handleTriStateChange = (event) => {
```
falconx, line 101: Moved logic to https://github.com/neontribe/gbptm/pull/588/files#diff-57fe27eb6d6dcf3d0d75c740b0a0f82eR38

src/components/EntryForm.js
```diff
@@ -0,0 +1,359 @@
...
+            (saveResponse && saveResponse.submitReport.code !== '200')) && (
+            <Notification>
+              Oops, there was an error saving your changes.
+              {console.log(saveError, saveResponse)}
+            </Notification>
+          )
+        }
+
+        <div className={controls.btnStack}>
+          {isFunction(children) ? children({ dirtyFields }) : children}
```
falconx, line 300: Allow children to be render prop to set `disabled` on the `EditPage` when no fields changed.
Intended to replce [onlyChanges](https://github.com/neontribe/gbptm/pull/588/files#diff-332fe77b4fa1c3806c14e30751be664fL547) :crossed_fingers:

src/pages/AddPage.js
falconx: Consolidated logic with https://github.com/neontribe/gbptm/pull/588/files#diff-b3d3457ad483621b96510b33c571eebfL153

stphnnnn:
This looks so much clearer, nice one. A couple of bugs:

1. If you edit a toilet (https://gbptm-git-form.neontribe.now.sh/loos/c36ddb71db1202ad5cf28c47/edit) and then move the map, the lat and lng fields change values correctly but the submit button does not become enabled as it should

2. Trying to add a toilet results in some GraphQl errors:
Screenshot 2020-04-29 at 20 54 32
I only added a title to the toilet so it may be a problem with the default values

falconx:
@stphnnnn (2) other than the `$location` error (the first one listed there, which I've now resolved), all these errors seem to also occur on the base branch.

PR #617, opened 2020-05-04
test: setup jest and testing-library

### LINKED ISSUE #609: react-testing-library setup
https://testing-library.com/docs/react-testing-library/intro

falconx:
Fixes #609

### DIFF
--- a/src/setupTests.js
+++ b/src/setupTests.js
@@ -1,0 +1,1 @@
+import 'mutationobserver-shim';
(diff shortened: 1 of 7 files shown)


src/setupTests.js
```diff
@@ -0,0 +1 @@
+import 'mutationobserver-shim';
```
falconx, line 1: https://github.com/react-hook-form/react-hook-form/blob/master/setup.ts

PR #625, opened 2020-05-05
build: add emotion and styled-system

### LINKED ISSUE #624: emotion + styled-system setup

falconx:
Fixes #624

Wasn't so sure about this https://emotion.sh/docs/install#with-babel-plugin-emotion, so I haven't included it in this PR.

> If you’re using Create React App, you can use the Babel macro

> There are some optimisations which aren’t possible with Babel Macros, so if it’s possible, we still recommend using babel-plugin-emotion


config-overrides.js
```diff
@@ -1,13 +1,24 @@
...
 
 module.exports = {
-  webpack: function (config, env) {
-    config.plugins.push(
-      new webpack.ProvidePlugin({
-        L: 'leaflet',
-      })
-    );
+  webpack: override(
+    useBabelRc(),
```
falconx, line 11: add `.babelrc` to be read

config-overrides.js
```diff
@@ -1,13 +1,24 @@
...
+    useEslintRc(),
+    addBabelPreset('@emotion/babel-preset-css-prop'),
```
falconx, line 13: allow `css` prop to work, e.g:
```
import { css } from '@emotion/core';
const Example = () => <div css={css`color: hotpink;`}>Example</div>
```

config-overrides.js
```diff
@@ -1,13 +1,24 @@
...
+          L: 'leaflet',
+        })
+      );
 
-    return config;
-  },
+      return config;
+    }
+  ),
 };
```
stphnnnn, line 24: I wonder if we should do away with react-rewired- app entirely to reduce the number of dependencies and deviation from a standard create react app application.

We could use the:
- JSX pragma when we need to use the css prop (unless styled-system relies on it?)
- It looks like with import leaflet everywhere it's needed so we don't need the ProvidePlugin for that

What do you think @falconx?

marxian:
I'd love to do away with the rewire, but I think we need it to apply some
extra webpack to leaflet, which is needed because leaflet and Adobe
tracking compete for the same global.

Insert vomit emoji here

config-overrides.js
falconx: I'm hoping it's not necessary to use the CSS prop, so I almost didn't add it.

We could use the JSX pragma and it's not a requirement for styled-system however it does mean their own [css prop documentation](https://styled-system.com/css) would only half work too — in this example, without `@emotion/babel-preset-css-prop`, the `Boop` component would be valid, but `Beep` would not.

So whilst I imagine we could get away without it, given that it looks like the config overrides are here to say anyway, it might be worth keeping? I'm also thinking, even if you and I don't make use of it (and I haven't used either enough to know this likelihood), it could be a pain if anyone comes onto the project and expects those features to work — which would save us documenting that they don't. That said, I'm happy to go with a philosophy of removing and only adding when/if we find it necessary :)

stphnnnn: yup agreed, let's keep it in 👍

PR #628, opened 2020-05-06
feat: add react-styleguidist

falconx:


src/components/Box.js
```diff
@@ -3,7 +3,7 @@ import { compose, space, color, layout, flexbox } from 'styled-system';
 import { createPropTypes } from '@styled-system/prop-types';
 
 // https://styled-system.com/guides/build-a-box
-const Box = styled.div`
+export const Box = styled.div`
   ${compose(space, color, layout, flexbox)}
 
   // ensures the Box can shrink below its minimum content size when used as a flex item
```
stphnnnn, line 9: Looks like this is a superfuious Box component?

src/components/Box/Box.js
```diff
@@ -0,0 +1,22 @@
...
+`;
+
+Box.propTypes = {
+  ...createPropTypes(space.propNames),
+  ...createPropTypes(color.propNames),
+  ...createPropTypes(layout.propNames),
+  ...createPropTypes(flexbox.propNames),
+};
+
+export default Box;
```
stphnnnn, line 22: Don't think we need the named export and can just move the `/** @component */` above `export default...`

I tested this out locally and it made the `Warning: src/components/Box/Box.js matches a pattern defined in “components” or “sections” options in your style guide config but doesn’t export a component.` warning dissapear

src/components/LooMap/LocateMapControl.js
```diff
@@ -4,7 +4,8 @@ import { MapControl, withLeaflet } from 'react-leaflet';
 
 import styles from '../css/loo-map.module.css';
 
-class LocateMapControl extends MapControl {
+/** @component */
+export class LocateMapControl extends MapControl {
```
stphnnnn, line 8: I'm not sure if we should expose leaflet components to styleguidist as I think they will only render when they're a child of the map component.

src/components/Box.js
falconx: In that we only use the component once or because I forgot to delete this file when cloning to `Box/Box.js`?

src/components/Box/Box.js
falconx: Interesting that I don't get the warning locally. But yes, we can move the docgen comment above the default export to the same effect — I'm still in the habit where older styleguidist required both exports to support outputting the styles and prop type documentation correctly, but that doesn't appear to be an issue now :)

src/components/LooMap/LocateMapControl.js
falconx: Yeah I don't think so either. Other than `Box` (which was mainly introduced to demonstrate `styled-system` and `emotion` working), the style guide is basically blank. I suspect, for example, we also won't want to see `EntryForm`, `MarkerClusterGroup`, and `ProtectedRoute`,

src/components/Box.js
stphnnnn: the second one 😉

src/components/LooMap/LocateMapControl.js
stphnnnn: aha all good, i suppose we can blacklist components we don't want to appear somehow?
not important right now

src/components/Box.js
falconx: :sweat_smile: thanks.

src/components/LooMap/LocateMapControl.js
falconx: Yeah we totally can :slightly_smiling_face: I already added bunch to the `ignore` list in `styleguide.config.js`

PR #629, opened 2020-05-06
feat: new search component

stphnnnn:
Adds a custom geolocation search combobox which looks uses Google Maps API to return suggestions based on a search string as the user types.

It looks like this:
Screenshot 2020-05-06 at 12 17 51


src/components/LocationSearch/location-search.module.css
```diff
@@ -0,0 +1,69 @@
+.wrapper {
+  box-shadow: 0 2px 4px 2px rgba(0, 0, 0, 0.1);
+}
+
+.label {
+  /* Visually Hidden */
```
falconx, line 6: we have a `helpers.module.css` class `.visuallyHidden` — not that it really matters, since we'll be throwing all that away.

src/components/LocationSearch/LocationSearch.js
```diff
@@ -0,0 +1,103 @@
...
+            className={styles.input}
+            style={{
+              borderRadius: isOpen ? '5px 5px 0 0' : 5,
+              borderBottom: isOpen ? '1px solid #ccc' : 'none',
+            }}
+            {...getInputProps()}
+          />
+        </div>
+
+        <div {...getMenuProps()}>
```
falconx, line 67: * `getMenuProps` will apply `role="listbox"`
* `getItemProps` will apply `role="option"`

I think that the `option` role is required to be a direct descendant of `listbox`.

I'd recommend pairing this with a list html, like so:
```jsx
<ul {...getMenuProps()}>
  <li {...getItemProps({ item, index })} />
</ul>
```

stphnnnn: Great spot, thank you 🙇

src/components/LocationSearch/location-search.module.css
stphnnnn: I've left this for now as we'll be rewriting with Emotion and have a nice component for this :)

falconx: exactly :)

PR #638, opened 2020-05-07
feat: introduce filters page

falconx:
filters


stphnnnn:
Sort of minor, but I'd expect to see the active state to be on the right side of the switch https://material-components.github.io/material-components-web-catalog/#/component/switch

falconx:
Yep, both would be valid approaches https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role#Result

Will be easy enough to change once we see the designs.

PR #640, opened 2020-05-07
Remove nearest loos list

stphnnnn:
Slowly moving towards the updated UI...

- removes nearest loos list
- removes nearest loos number labels
- removes map center for all maps except edit and add pages


stphnnnn:
i will include these changes in another PR

PR #644, opened 2020-05-11
feat: add Header and Footer

falconx:
header-footer


src/components/PageLayout.js
```diff
@@ -127,46 +126,44 @@ const ResetStyles = (
...
-  React.useEffect(() => {
-    if (mainRef.current && isCookieSettingsOpen) {
-      mainRef.current.scrollTop = 0;
-    }
-  }, [mainRef, isCookieSettingsOpen]);
+  // React.useEffect(() => {
+  //   if (mainRef.current && isCookieSettingsOpen) {
+  //     mainRef.current.scrollTop = 0;
+  //   }
+  // }, [mainRef, isCookieSettingsOpen]);
```
falconx, line 139: Typically wouldn't leave comments in a PR, but felt useful to leave this stuff?

src/theme.js
```diff
@@ -1,6 +1,9 @@
 const theme = {
   colors: {
-    primary: 'hotpink',
+    primary: '#0a165e',
+    secondary: '#92f9db',
+    white: '#fff',
+    grey: '#f4f4f4',
```
falconx, line 6: We'll probably change these keys around once we have Zeplin to refer to.

src/components/PageLayout.js
stphnnnn: probably not useful anymore but we can remove later

PR #645, opened 2020-05-11
feat: support conditional viewport-based UI

falconx:


package.json
```diff
@@ -30,6 +30,7 @@
     "@apollo/client": "3.0.0-beta.44",
     "@apollo/link-context": "2.0.0-beta.3",
     "@apollo/link-error": "2.0.0-beta.3",
+    "@artsy/fresnel": "1.1.0",
```
marxian, line 33: Lovely package name!

falconx:
@marxian 

> Do we have aspects of the intended design which would need us to otherwise conditionally render?

I like their approach too! I'm currently working on the header which, for mobile only, displays a drawer. I'm hoping that we can use the library sparingly though.

PR #653, opened 2020-05-13
feat: filters

falconx:
* Search results still need styling (I don't have an API key)
* I think we'll want to make the drawers modal-dialogs — but this is a confusing topic so I've reached out to Heydon. I'm expecting we may need to tweak the design a little.
* General a11y improvements still to be made/reviewed
* Currently no drawer slider animation
* Main navigation drawer appears behind filter drawer


src/components/Button.js
```diff
@@ -13,8 +13,9 @@ const StyledButton = styled.button(
     display: inline-flex;
     align-items: center;
     justify-content: center;
-    user-select: none;
-    height: ${BUTTON_HEIGHT}px;
+    box-sizing: border-box;
```
stphnnnn, line 16: Do we need border-box here? Isn't these covered by the global reset?

I would recommend adding "user-select: none" some touch screen devices make it easy to select the button text when the user is trying to press the button. Unless there's a good use case for a user wanting to select the button text?

src/components/Switch.js
```diff
@@ -0,0 +1,45 @@
...
+import Box from './Box';
+
+const HEIGHT = 16;
+const WIDTH = 27;
+const OFFSET = 1;
+const LENGTH = HEIGHT - OFFSET * 2;
+
+const Inner = styled(Box)`
+  transition: left 0.5s ease;
+`;
```
stphnnnn, line 14: I'd recommend decreasing the animation length to 200ms or 300ms.
500ms is quite slow for this type of animation and could make the UI feel less responsive to users.

Really good article on the subject here: https://uxdesign.cc/the-ultimate-guide-to-proper-use-of-animation-in-ux-10bd98614fa9

src/components/Button.js
falconx: > Do we need border-box here? Isn't these covered by the global reset?

Yeah unfortunately the `all: unset` rule undoes it. On my personal projects I've gone off this rule and instead just remove `border` and `background` etc.

> I would recommend adding "user-select: none"

Oops, wasn't meant to remove :sweat_smile: I had crazy merge conflicts going on all over the place :grimacing:

PR #661, opened 2020-05-14
fix: hover/focus styles

falconx:


src/components/Button.js
```diff
@@ -57,11 +67,7 @@ const ButtonIcon = ({ icon }) => {
 
 const Button = ({ children, icon, ...props }) => (
   <StyledButton type="button" {...props}>
-    {Boolean(icon) && (
-      <Text color="primary">
-        <ButtonIcon icon={icon} />
-      </Text>
-    )}
+    {Boolean(icon) && <ButtonIcon icon={icon} />}
```
falconx, line 70: Removed because I needed the icons to inherit the current colour on hover/focus. Will this case a problem anywhere for you @stphnnnn?

stphnnnn: don't think so

PR #666, opened 2020-05-15
feat: tracking banner

falconx:


falconx:
* I agree there's too much prop drilling right now. I believe composition over context would be preferenced here. In each case we're trying to get a couple of props to a single control which could be defined higher up.
* I'm not sure about the reloading and wouldn't know how to test a refactor.

PR #669, opened 2020-05-15
feat: use new map markers

stphnnnn:


marxian:
Yay! I've been looking forward to this.

Is it worthwhile considering an outline to help when the markers stack, especially when we're no longer clustering?

stphnnnn:
> Yay! I've been looking forward to this.
> 
> 
> 
> Is it worthwhile considering an outline to help when the markers stack, especially when we're no longer clustering?

Yeah I think so 👍

PR #673, opened 2020-05-16
Migrate loo props

marxian:


src/api/db/manage/generate/index.js
```diff
@@ -0,0 +1,61 @@
...
+      if (e.code !== 26) {
+        throw e;
+      }
+    }
+
+    // create set of new loos from new reports
+    console.warn('Migrating loos across from new reports');
+    const newLoos = await generateFromReports();
+
+    console.error(newLoos.length + ' loos generated');
```
falconx, line 50: Should this be an "error" or just a log?

PR #689, opened 2020-05-19
fees migrate

marxian:


src/components/EntryForm.js
```diff
@@ -195,32 +200,32 @@ const EntryForm = ({
...
             This toilet is free
             <input
               ref={register}
               name="isFree"
               type="checkbox"
-              checked={isFree}
+              checked={!isPaymentRequired}
               className={styles.feeToggle}
-              onChange={(event) => setIsFree(event.target.checked)}
+              onChange={(event) => setIsPaymentRequired(event.target.checked)}
```
stphnnnn, line 213: I think `event.target.checked` needs to be inverted here.
If the checkbox is checked then we should set `isPaymentRequired` to `false` instead of `true`

marxian: I think so too.

I got tangled up in booleans

PR #700, opened 2020-05-21
fix: IE11 + Safari

falconx:


src/components/Footer.js
```diff
@@ -34,7 +34,7 @@ const Footer = ({ children }) => (
     py={[0, 2]}
     bg={['transparent', 'lightGrey']}
     color="primary"
-    minHeight={60}
+    height={['auto', 60]}
```
falconx, line 37: Fix IE11 issue with vertically centering in flexbox.

src/components/Header/Logo.js
```diff
@@ -9,7 +9,7 @@ const Logo = (props) => (
     as="img"
     src={logo}
     alt="The Great British Toilet Map"
-    height={[30, 36]}
+    width={154}
```
falconx, line 12: Avoid IE11 issue scaling logo image. I tried the min-height/width fix on the flexbox item but couldn't get anything else to work.

src/components/PageLayout.js
```diff
@@ -84,10 +85,19 @@ const ResetStyles = (
       /* have link and buttons be indistinguishable */
       a,
       button {
-        all: unset;
```
falconx, line 87: I'm gonna stop using this in resets.

src/components/PageLayout.js
```diff
@@ -200,12 +210,12 @@ const PageLayout = (props) => {
         <Box display="flex" flexDirection="column" height="100%">
           <Header>{footerFragment}</Header>
 
-          <Box position="relative" display="flex" flexGrow={1}>
+          <Box position="relative" display="flex" flexGrow={1} overflowY="auto">
```
falconx, line 213: I'd like to get this approach approved with everyone first. It would prevent the IE11 footer being positioned on top of the content for long pages.

(which happens because we `height: 100%` all containers to allow the footer to `margin-top: auto` itself and therefore be positioned at the bottom of the screen for short pages. min-height: 100%/vh also do not solve this issue.)

stphnnnn: This should be achievable with flex unless I'm missing something?
What if we add `alignItems="stretch"` to the wrapping Box?

falconx: This won't work. I don't believe in this case there's a fix which would satisfy all browsers. Feel free to play though.

PR #713, opened 2020-05-26
fix: make map keyboard accessible

falconx:
1. Added focus styles to map container and markers.

2. Added `role="application"` to map container.

NVDA (the only screen reader I've tested in) was announcing the entire `role="main"` region contents when focusing the map container. Screen readers also override keys, making it not possible to pan the map without the `application` role.

I discovered [this very active issue](https://github.com/Leaflet/LEaflet/issues/3210#issuecomment-258918073) where someone else has reached the same conclusion. Though I'm still not entirely comfortable with applying this role. I haven't found a way to tab past or escape the map region once it has been entered, other than by tabbing through the entire marker collection — since whilst the user is focused within the application role, a screen reader user cannot jump between headings or landmarks.

We could remove `tabindex="0"` from the map container, but this would prevent keyboard users from panning the map.

3. Added `role="link"` to markers, labelled by the toilet name or "Unnamed toilet" if no name is available.

Not sure what else we could use as a fallback here.

`<Link>` not supported by `react-leaflet`.

4. When selecting a marker I'd like to programmatically set the focus to the heading within the `ToiletDetailsPanel`. I haven't found a way to achieve this yet whilst also not updating focus under other circumstances (such as clicking on the "More details" control)


src/components/LooMap/LooMap.js
```diff
@@ -56,11 +74,18 @@ const LooMap = ({
...
-              push('/loos/' + toilet.id);
+              push(`/loos/${toilet.id}`);
+            }
+          }}
+          onKeyDown={(event) => {
+            if (!staticMap && event.originalEvent.keyCode === KEY_ENTER) {
+              push(`/loos/${toilet.id}`);
             }
           }}
+          keyboard={!staticMap}
```
falconx, line 88: From the Leaflet API. Removes `tabindex="0"` from markers.

src/components/LooMap/Marker.js
```diff
@@ -0,0 +1,23 @@
...
+
+const CustomMarker = ({ label, ...props }) => {
+  const markerRef = useRef(null);
+
+  useEffect(() => {
+    const marker = markerRef.current.leafletElement.getElement();
+
+    marker.setAttribute('role', 'link');
+    marker.setAttribute('aria-label', label);
+  });
```
stphnnnn, line 14: Can we add an empty dependency array here too?

src/components/LooMap/LooMap.js
```diff
@@ -23,11 +27,25 @@ const LooMap = ({
...
+    // without the application role screen reader software overrides these controls
+    // this also avoid the entire main region being announced
+    map.setAttribute('role', 'application');
+    map.setAttribute('aria-label', 'use arrow keys to pan the map');
+    map.setAttribute(
+      'aria-keyshortcuts',
+      'ArrowUp ArrowDown ArrowLeft ArrowRight'
+    );
+  });
+
```
stphnnnn, line 46: I think we will only need to apply these on the first render (unless they are overidden?) in which case a empty dependecy array should be passed as the second argument to useEffect

PR #719, opened 2020-05-27
Fix getMenuProps downshift error

stphnnnn:
- fixes `getMenuProps` downshift error

### DIFF
--- a/src/components/Sidebar.js
+++ b/src/components/Sidebar.js
@@ -35,6 +35,11 @@ Arrow.propTypes = {
   isExpanded: PropTypes.bool,
 };
 
+const StyledNavLink = styled(NavLink)`
+  display: flex;
+  align-items: center;
+`;
+
 const Sidebar = ({
   filters,
   onFilterChange,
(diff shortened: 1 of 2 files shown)


src/components/Sidebar.js
```diff
@@ -115,11 +120,8 @@ const Sidebar = ({
         <h2 id="heading-add">
           <VisuallyHidden>Add a Loo</VisuallyHidden>
         </h2>
-        <Box
-          as={NavLink}
+        <StyledNavLink
```
falconx, line 123: out of interest, was `as={NavLink}` causing problems or was it changed due to preference?

stphnnnn: it was causing problems, the alignItems prop was being applied as an attribute and my browser was complaing

stphnnnn:
> 
> 
> \o/ and fixes an issue where the listbox was referencing a missing id.
> 
> (minor) there is still an issue where we have no results, since the `listbox` role still expects direct children to be role `option`.

I thought this too but I couldn't actually find that specified anywhere. Either way, it feels like a small(ish) concession.

falconx:
> I thought this too but I couldn't actually find that specified anywhere. Either way, it feels like a small(ish) concession.

You can see the issue by displaying no results and running the axe developer tools. But yeah it's definitely minor.

PR #743, opened 2020-06-01
fix: improve map accessibility

falconx:
* Would have preferred the map controls to be before the map in tab order. This would allow the details panel to be the next tab stop following the map
* Might be nice to show the numbers on the map too

Screenshot from 2020-06-02 12 25 18


marxian:
@falconx After tabbing into the map, and then using number keys to select some loos I experience mutiple keydown handler violations.

The more loos I visit using keystrokes the more violations stack, 'till my browser becomes unresponsive.

falconx:
@marxian Thanks. Now that I've tested properly, I also now see that the "6" key causes the map to zoom out.

stphnnnn:
Could a different set of keys be used for to navigate between markers to get around the "6" key problem?

src/components/LooMap/LooMap.js
```diff
@@ -85,12 +93,34 @@ const LooMap = ({
...
+    };
+
+    const observer = new MutationObserver(callback);
+
+    // only render accessibility overlays when [data="focus-visible-added"] is applied
+    //
+    // we conditionally render instead of toggling CSS display since we want to avoid AccessibilityList being announced
+    // before the map is keyboard focused
+    observer.observe(map, { attributes: true });
+  }
```
stphnnnn, line 122: Creating a mutation observer is a side effect of the component so should probably go inside a useEffect with a disconnect cleanup

src/components/LooMap/AccessibilityIntersection.js
```diff
@@ -0,0 +1,65 @@
...
+
+      onSelection(selectionIndex);
+    };
+
+    document.addEventListener('keydown', keyDownHandler);
+
+    return () => {
+      document.removeEventListener('keydown', keyDownHandler);
+    };
+  });
```
stphnnnn, line 43: Currently this will tear down existing event handlers and create new ones every render, we probably want an empty dependency array here.

You might also have to wrap the onSelection function in useCallback wherever that is defined

PR #744, opened 2020-06-01
Bundle size reduction

stphnnnn:
- replaces luxon with date-fns
- code splits some routes
- removes some unused dependencies
- moves leaflet.css import to the loomap component
- refactors the auth provider to be more inline with auth0 and react docs


package.json
```diff
@@ -79,10 +80,10 @@
     "react-hook-form": "5.7.2",
     "react-leaflet": "2.7.0",
     "react-leaflet-control": "2.1.2",
-    "react-responsive": "8.0.3",
     "react-router-dom": "5.2.0",
     "react-router-hash-link": "1.2.2",
     "resize-observer-polyfill": "1.5.1",
+    "source-map-explorer": "2.4.2",
```
marxian, line 86: Dev dependency please!

PR #756, opened 2020-06-03
feat: display use our loos campaign toilets

falconx:
uol-pins

uol-sm

uol-md


src/components/EntryForm.js
```diff
@@ -509,12 +522,68 @@ const EntryForm = ({ title, loo, center, children, ...props }) => {
...
+                }}
+              />
+
+              <Spacer ml={3} />
+
+              <Text as="p" fontSize={2}>
+                Use our loos is a national campaign working to unlock the loos
+                that are hidden in high street cafes, restaurants, coffee shops
+                and bars.&nbsp;
+                <Button
```
marxian, line 571: @falconx  Maybe we should send to target=_blank to avoid losing part-filled form data?

falconx: oh yes, definitely.

PR #757, opened 2020-06-03
fix: sidebar

falconx:
* allow pins below the sidebar to be selected on larger viewports
* avoid details panel overlapping with the sidebar, plus a little extra spacing so they don't touch


src/pages/HomePage.js
```diff
@@ -20,6 +20,8 @@ import config, { FILTERS_KEY } from '../config';
 
 const FIND_BY_ID = loader('./findLooById.graphql');
 
+const SIDEBAR_BOTTOM_MARGIN = 32;
```
falconx, line 23: Should ideally use `theme.space` but `useTheme` doesn't work directly in this component and decided not to spend ages investigating.

PR #760, opened 2020-06-04
feat: add styleguide

falconx:
Probably not a good idea to merge this before a transition to NextJS. I can port this code afterwards if that does go ahead.

I'm also keeping an eye on https://github.com/reactjs/react-docgen/pull/352 which will allow for the styled system props to be documented.

screencapture-localhost-6060-2020-06-04-11_22_11


marxian:
@falconx I'm not clear about whether to merge this at this point?

falconx:
@marxian I'm happy for it to be merged. I just didn't want it to interfere with the transition nextjs. But, since that work is not happening right now, there shouldn't be a problem.

PR #771, opened 2020-06-04
replace Apollo Client with SWR

stphnnnn:
- apollo local schema has been replaced with react context state
- apollo queries (except in explorer) have been replaced with SWR
- apollo mutations
- replace apollo queries in explorer
- remove apollo things
- adds fragments to make our cache mutations more predictable


src/pages/EditPage.js
```diff
@@ -17,42 +18,62 @@ import LooMap from '../components/LooMap';
...
-  const { loading: loadingLooData, data: looData, error: looError } = useQuery(
-    FIND_BY_ID,
-    {
-      variables: {
-        id: props.match.params.id,
-      },
-    }
+  const { loading: loadingLooData, data: looData, error: looError } = useSWR(
+    `{
+      loo(id: "${props.match.params.id}") {
```
marxian, line 31: This makes loads of alarms ring in my head.

Can we pass variables? https://github.com/prisma-labs/graphql-request#using-variables

stphnnnn: yup i think that looks nicer, although the results are the same; a unique cache key for each find loo by id query

marxian: For me it triggered old "passing user input from the URL into a database query by concatenation" reflexes

stphnnnn: aha makes sense

src/pages/EditPage.js
```diff
@@ -16,43 +16,40 @@ import EntryForm from '../components/EntryForm';
...
+  ]);
 
-  const [mapPosition, setMapPosition] = useMapPosition();
+  const [mapState, setMapState] = useMapState();
 
   const looLocation = (looData && looData.loo.location) || null;
 
   // set the map position to the loo location
   React.useEffect(() => {
+    console.log('setting loo location', setMapState, looLocation);
```
falconx, line 43: log.

jthistle:
I'm confused: first we moved away from Redux, because we didn't like Redux, and now we're moving back to something that's like... Redux? Or maybe I just don't understand React Context State 😅

marxian:
@jthistle I can see the cause for confusion :-)

I can also see it might be surprising to see your excellent local state work in apollo fade away (though there was good value learning there for this project).

Truth is that the nature of the data (and the way we use it) means that we get very little extra value from the main Apollo Client cache, and that became increasingly evident when we assessed the real cost of that package in terms of bundle size.

Taking that line, of course, removes the ability to keep local state data in the Apollo Client and we have to put it somewhere. You're right to see the use of context, a hook and useReducer as being a lot like redux (though with a lot less ceremony).

This move is also intended to smooth the way to a transition to Next.js. We can integrate Next and Apollo Client - we've done it a lot in fact - but it is fiddly and doesn't bring much benefit for this app.

jthistle:
Got it, thanks. I'm not worried about my code just disappearing, that's the very nature of the stuff. Just slightly bamboozled about the decision, although not any more

marxian:
> I'm not worried about my code just disappearing, that's the very nature of the stuff. 

That is a proper truth and an excellent turn of phrase.

PR #804, opened 2020-06-10
feat: reinstate voyager

marxian:


jthistle:
I presume we import the Voyager at some point into the explorer to mount it? Or does that just magically happen?

marxian:
Blimey! I'd forgotten some stuff. Thanks @jthistle!

PR #824, opened 2020-06-17
fix: coherent men/women/children

marxian:


harryharrold:
(Aside: best fixname ever.)

PR #863, opened 2020-07-08
feat: show covid notification

falconx:
covid-notification-blue-1

covid-notification-blue-2


marxian:
@falconx I'm seeing a different colour banner than in your screenshots. Is that intentional?

falconx:
@marxian I think Github must be cached for you. I'd switched to the primary blue and updated the description. This matches your screenshot though so we're all good :)

PR #882, opened 2020-07-21
feat: use nominatim geocoder for search

stphnnnn:
Something's not working with the Google Places session tokens which is making the cost unsustainable. This PR reverts to using the Nominatim geocoder for our search results.

- adds a `useNominatimAutocomplete` hook to fetch search results based on the query input
- swaps `useNominatimAutocomplete` for `usePlacesAutocomplete`
- tweaks `usePlacesAutocomplete` to make the API more similar to `useNominatimAutocomplete` and generic which should mean future search provider changes will not require changes to the `LocationSearch` component itself

### DIFF
--- a/src/components/LocationSearch/useNominatimAutocomplete.js
+++ b/src/components/LocationSearch/useNominatimAutocomplete.js
@@ -6,7 +6,7 @@ const useNominatimAutocomplete = (input) => {
 
   const fetchPlaces = React.useCallback(
     debounce(async (input) => {
-      const fetchUrl = `https://nominatim.openstreetmap.org/search?q=${input}&limit=5&format=json`;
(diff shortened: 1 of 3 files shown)


src/components/LocationSearch/useNominatimAutocomplete.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import debounce from 'lodash/debounce';
+
+const useNominatimAutocomplete = (input) => {
+  const [places, setPlaces] = React.useState([]);
+
+  const fetchPlaces = React.useCallback(
+    debounce(async (input) => {
+      const fetchUrl = `https://nominatim.openstreetmap.org/search?q=${input}&limit=5&format=json`;
```
marxian, line 9: This seems to work nicely.

I wonder if we can decrease the noise in the results by adding `countrycodes=gb` as seen here https://nominatim.org/release-docs/develop/api/Search/#result-limitation

Issue #1006, opened 2021-01-20
Won't update Opening Hours 


gailramster:
When editing a toilet page, if I enter opening hours, the 'Update the toilet' button does not do anything.
If I make edits but leave opening hours as unknown, it updates.

is there a particular format that these need entering in, to make it work? 

Using Safari on desktop. 

(it was this toilet https://www.toiletmap.org.uk/loos/27c1892458a07c7e7e3089aa - I've abandoned edits before when they wouldn't submit, I think for same reason, but just pinpointed the opening hours as likely problem)

marxian:
@gailramster I found some issues in opening hours, and I've deployed a new release.

However, I'm interested i your question "is there a particular format that these need entering in, to make it work?"

Do you not see a pair of 24hour time pickers for the opening and closing time on each day?

gailramster:
Ah! No. 

Safari 13.1.2

![Screen Shot 2021-02-01 at 12 38 31](https://user-images.githubusercontent.com/2690575/106460089-b501f180-648a-11eb-9d6a-afb633e609da.png)

I've just checked on Chrome and can see them now - but I always use Safari..

marxian:
Fascinating :-)

I'll look into it :-)

PR #1010, opened 2021-02-01
fix: only allows number input

roseb04:
Fixing this issue: https://github.com/neontribe/gbptm/issues/1006

### DIFF
--- a/src/components/EntryForm.js
+++ b/src/components/EntryForm.js
@@ -490,7 +490,7 @@ const EntryForm = ({ title, loo, center, children, ...props }) => {
                           <input
                             type="time"
                             pattern="[0-9]{2}:[0-9]{2}"
-                            placeholder="08:00"
+                            placeholder="hh:mm"
                             defaultValue={
                               loo.openingTimes
                                 ? loo.openingTimes[index][0]
@@ -511,7 +511,7 @@ const EntryForm = ({ title, loo, center, children, ...props }) => {
                           <input
                             type="time"
                             pattern="[0-9]{2}:[0-9]{2}"
-                            placeholder="24:00"
+                            placeholder="hh:mm"
                             defaultValue={
                               loo.openingTimes
                                 ? loo.openingTimes[index][1]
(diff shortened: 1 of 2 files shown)


roseb04:
Fix doesn't currently let users know when they are incorrectly typing in a date until they've clicked submit. Even if they do put in the number format then they could put in a nonsensical time eg 25:99.

This article might hold some clues on how to make a bit of a clever fix - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#handling_browser_support

marxian:
@gailramster when you get a moment please have a little test of @roseb04's interim solution here (it should solve the worst of the problem): https://gbptm-git-fix-opening-hours-safari.neontribe.vercel.app

gailramster:
Looks like a good solution thought it's not letting me edit the numbers in Safari.

Screen Shot 2021-02-01 at 19 46 35

marxian:
@gailramster We've done some investigation - it is certainly a standards-compliance issue in Safari (that's not unusual). Things are much better in current Safari editions but you seem to be pinned in an older release of OSX. Estimates suggest that 

We have options still, to fix this but it adds unwelcome complexity. @rose is thinking to add a format indicator...

roseb04:
@gailramster I've just checked with a colleague who has the same Safari browser that you are using. It appeared that while he couldn't edit the numbers (like you could in Chrome), he could type in the box and edit that way. Are you able to type in the time?

gailramster:
@roseb04 Hi Rose, yes now I know the 00:00 format, I can add times and they'll submit, on the toiletmap.org.uk site. I can't edit the pre-filled numbers on the interim solution.  Would suggesting the format somewhere in the question be enough?  I was probably writing 0:00, 0.00 or 0pm.

roseb04:
@gailramster that is great news. I've updated the placeholder text so that is should be clearer that the format should be hh:mm. I will get that approved and merged in, so we can close this issue.

marxian (approved):

Issue #1032, opened 2021-03-25
Make it easier for folk to know that they can geolocate


marxian:
**Is your feature request related to a problem? Please describe.**
It seems that we're not sure if people know they can zoom the map to their location. Maybe they don't find the map control.
Seems that people use the search
**Describe the solution you'd like**
Probably a more prominent control to invoke geolocation

**Describe alternatives you've considered**
We tried geolocating on load but it is annoying and against the W3C guidelines

**Additional context**
Maybe we should gather info now we've got plausible before we commit to a course of action?

marxian:
I think the first thing to do it check the stats here: https://plausible.io/share/toiletmap.org.uk?auth=BW3helL2W3OVESMSObS_M

Then maybe instrument the search and the geolocate controls using a plausible event (https://plausible.io/docs/custom-event-goals) 

I've set up two events as goals `Geolocate` and `Search`

PR #1049, opened 2021-04-07
feat: add plausible events

rosejbon:
Adding plausible events to see how many users are searching or using geolocate to find toilets.

Related to issue #1032. We want to double check our assumption that users are finding it hard to find toilets near them, and aren't using geolocate.

### DIFF
--- a/src/components/LooMap/LocateMapControl.js
+++ b/src/components/LooMap/LocateMapControl.js
@@ -58,11 +58,20 @@ const LocateMapControl = ({ position }) => {
     onStopLocation,
   });
 
+  const handleClick = () => {
+    if (!isActive) {
+      startLocate();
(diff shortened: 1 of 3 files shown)


marxian (approved):

src/components/LooMap/LocateMapControl.js
```diff
@@ -58,11 +58,20 @@ const LocateMapControl = ({ position }) => {
     onStopLocation,
   });
 
+  const handleClick = () => {
+    if (!isActive) {
```
ob6160, line 62: not a huge deal really, but swapping the condition so we're not using the not operator (like the condition was before the change) might help with readability

ob6160 (approved):
looks good!

Issue #1037, opened 2021-03-29
Cannot close a toilet on a certain day (opening hours)


gailramster:
The toggle switch to turn a toilet from 'closed' to 'opening hours' on a certain day, does not turn off again. 

I've tried deleting the numbers in the opening hours field, and clicking 'update toilet' to try to close a toilet on that day. This doesn't load a new page to confirm that the toilet has been updated - it stays on the edit page. 

Though the toggle button looks squished on Safari, this happens on both Safari and Chrome. 

e.g. https://www.toiletmap.org.uk/loos/3d96d12c38bf7edb22624957/edit
This cafe at Westminster Central Hall is closed on Sat & Sun. I 'opened' it on Saturdays to test the opening hours, and now I cannot close it again!

![Screen Shot 2021-03-29 at 10 02 54](https://user-images.githubusercontent.com/2690575/112814666-7007cf00-9077-11eb-8712-78f40fd938f3.png)
![Screen Shot 2021-03-29 at 10 03 16](https://user-images.githubusercontent.com/2690575/112814676-7138fc00-9077-11eb-9eef-4ce812afc8bc.png)

ob6160:
hey Gail, will take a look!

PR #1072, opened 2021-05-01
fix: switch elements after react form update

ob6160:
Resolves https://github.com/neontribe/gbptm/issues/1037

Seems as though an update to react hook form at some point changed how state is managed on form controls

Also forwards ref to switches so react hook form focus behaves as expected

Not sure if we have an E2E test suite, but might be helpful in catching regressions like this


marxian (approved):

marxian:
Yay! Thanks @ob6160 

I feel bad because I'm pretty sure I let this one through. Thanks for covering for me.

PR #1080, opened 2021-05-07
feat: add find toilet near me button

rosejbon:
Part of #1032, adding a more visually obvious button for users on mobile to click and find loos near them. We will track it using Plauisble and compare to see if it works.


marxian:
Looks good, but, it seems to me that the new button is only visible on smaller viewports.  Is that the intention?

rosejbon:
> Looks good, but, it seems to me that the new button is only visible on smaller viewports. Is that the intention?

That is a good point Rupert. I was making an assumption about how people use the site. I will add a button onto the larger screen.

ob6160 (approved):
nice one

marxian (approved):
Nice! Good on big screens too now... And I can see the event getting sent.

PR #1085, opened 2021-05-09
Feat/groundwork tests

ob6160:
## Description
* Adds a new test suite which ensures that key loo edit form elements are behaving as expected. Prompted after switches were broken for a while on the edit page #1072 
  - Fixes a problem when running tests locally where puppeteer wasn't tearing down properly when tests fail midway through.. leaving the browser process hanging.
* Decided to chuck in a jsconfig file so we get some nice types for our Puppeteer tests :)
* Was unable to test a full form submission due to issues mocking the backend. Will explore this further in the future.


marxian (approved):
Lovely! This is reassuring!

(my only note of substance is that these commits might have been better prefixed with `test:`, but we're only using semantic versioning for form at the moment)

PR #1192, opened 2021-10-03
Ob/next edit page

ob6160:
combines changes from #1182 along with getting the loo edit page to work again


marxian (approved):
Nice, my part is pretty unfinished, but overall I think this constitutes an improvement ('specially your part) and we should merge it...

ob6160:
> Nice, my part is pretty unfinished, but overall I think this constitutes an improvement ('specially your part) and we should merge it...

Cool cool, I'm moving on to fixing the loo deletion flow, then I think almost all of the pages will be functional 🎉

PR #1203, opened 2021-10-14
fix: react-leaflet

marxian:
Load compressed data for loo markers (prototype)
Fill the apollo cache on the homepage, and try to use it thereafter.
Load the dataset directly when visiting a loo page cold.
Take steps to preserve apollo-client cache when navigating to pages which have their apolloState data

 - [x] Filtering is currently broken - I think we need to keep filters on a context and respect them in the Markers component.We also need to devise an encoding for the filterable properties.
- [x] The cache isn't quite working
- [x] The API should cache the compressed output
- [ ] Pages using SSG should set revalidate where appropriate
- [ ] Propagate the approaches in / and /loos/:id to other pages
- [x] Have a big debate about cache persistence/SSG/service-workers for the marker data


ob6160:
decided to merge and close this as it's getting quite unwieldy

PR #1258, opened 2021-12-04
Map filter refactor

### LINKED ISSUE #1250: Manage filters without prop-drilling

ob6160:
This allows us to use the useMapState hook to access filters whereever needed

This is instead of having to rely upon prop drilling to pass them down to the level that we would like.

Also includes some further improvements made in #1259 

Closes #1250


marxian (approved):
Lovely. So much neater.

PR #1275, opened 2021-12-11
feat: lifts the currently focused loo up to the map state

ob6160:
also removes the focused loo from the leaflet loo group so that the marker overlap issue is avoided

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "@mapbox/geojson-rewind": "0.5.1",
     "@types/jest": "27.0.3",
     "@types/leaflet": "1.7.6",
-    "@types/leaflet.markercluster": "1.4.6",
+    "@types/leaflet.markercluster": "^1.4.6",
     "@types/lodash": "4.14.177",
     "@types/node": "16.11.11",
     "@types/react": "17.0.37",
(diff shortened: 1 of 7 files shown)


marxian (approved):
I don't object to this, but I wonder if the tradeoff will prove to be worth it...

This certainly fixes the overlapping issue, but that only occured in the first place because I'd arbitrarily decreased the precision of marker locations for the cluster layer. If we matched precision (which I think we should anyway) there'd not have been an issue to fix.

I have a lurking suspicion that doing it this way will involve rerendering the markercluster layer more often, which could be expensive.

ob6160:
> 



> I don't object to this, but I wonder if the tradeoff will prove to be worth it...
> 
> This certainly fixes the overlapping issue, but that only occured in the first place because I'd arbitrarily decreased the precision of marker locations for the cluster layer. If we matched precision (which I think we should anyway) there'd not have been an issue to fix.
> 
> I have a lurking suspicion that doing it this way will involve rerendering the markercluster layer more often, which could be expensive.

I think with geohashes we can match the position now, this way does feel more "correct" though.. now that we're looking at chunking in #1276 I think the overhead re-rendering the marker layer won't be too bad because it'll just involve re-rendering each chunk , not the whole map. i think we should switch back to the other method though if performance does turn out to be a problem..

and a related though: it would be really nice if the marker got larger immediately upon click to indicate that you've selected a loo (even if it hasn't loaded yet).. not immediately sure how easy that'd be to do.

PR #1276, opened 2021-12-13
Geohash fun (WIP)

ob6160:
Implements a gql query & hook for querying toilets by their `geohash` value at different levels of precision. This lets us request toilets at different levels of detail! 

We load in chunks based upon the lat/lng centre view of the map (apart from at the national level, when it's centred on the UK). Each neighbour and their neighbours excluding duplicates are loaded in to grab a sensible portion of the map that covers the users' viewport. Each chunk is rendered separately, making use of the aforementioned geohash gql query hook

The precision of the geohash is changed based upon the zoom level, so that the chunks get larger as the user zooms out.

Here are some pictures of geohash chunking at each zoom level (blue zones turned on for debugging):

## National
Screenshot 2021-12-17 at 16 26 39

## Birds eye
Screenshot 2021-12-17 at 16 19 20

## Local
Screenshot 2021-12-17 at 16 19 32

## Street level
Screenshot 2021-12-17 at 16 33 05


marxian (commented):
Well this is awesome!

The client-side is unexpectedly neat.
The server side is very efficiently cacheable.

src/components/LooMap/Markers.tsx
```diff
@@ -25,9 +27,21 @@ const mcg = L.markerClusterGroup({
...
 
-  const [mapState, setMapState] = useMapState();
+  const [mapState] = useMapState();
   const { filters } = mapState;
-  const { data } = useUkLooMarkersQuery();
+
+  const map = useMap();
+  console.log(map.getZoom());
+
+  const hashPrecision = map.getZoom() < 11 ? 3 : 4;
```
marxian, line 36: Very interested in how this might scale across more zoom levels... Was this an initial proof-of-concept, or do you think this mapping might be sufficient?

ob6160: I've just pushed a big improvement to this mapping, it changes the geohash "chunk" precision according to the zoom level on a national, birds-eye and local basis. It also adds some debug squares so we can see how it's doing. I think we'll need to make some decisions/put more thought into how big the chunks are at the higher zoom levels so we can decide between bigger payloads vs. more requests for smaller chunks. 

I think this is definitely the way to go though, so simple to implement and very efficient!

ob6160:
@marxian eager to see what you think about the latest build :) https://gbptm-next-ionscniq8-neontribe.vercel.app/ – still needs tweaking, but I think the approach has definite promise! I've split the marker loading out into "chunks" and highlighted at each zoom level with a debug rectangle. Each chunk can load in as the network request completes, so we don't block and the user can see loos as soon as they are available

src/components/LooMap/Markers.tsx
ob6160: Screenshot 2021-12-17 at 16 15 01

This is one of those places where it'll be a trade-off between payload size and number of chunks. If I bumped up the precision at this national zoom level they'd be making a lot more requests to get all of the sub chunks... but also there wouldn't be that single one which contains most of the UK loos...

perhaps we could keep the chunk size large at this zoom level but better centre it over the UK so there's an even distribution inside 🤔

ob6160:
Another thought that just occurred, now that we have these zoom levels, we could extend the `stringifyAndCompress` method to also include enough data to show the toilet info panel without having to make extra network requests at the closer-in zoom levels maybe

PR #1289, opened 2022-02-12
Fix accessiblity overlay

### LINKED ISSUE #1273: Fix accessibility list on toilet map (next)

ob6160:
## What has changed?

Fixes @falconx's accessibility overlay so that it works correctly with the new toilet clustering & geohash loading infrastructure.

I don't 100% like how I've gotten this to work but now we're nearing (fingers crossed) the last few big parts of the nextjs migration the next big thing after launching will be to tidy up stuff like my half-baked implementation here.

deployment: https://gbptm-next-3k5ig1cyq-neontribe.vercel.app/

Used a slightly awkward way of recording which toilets have been loaded in already for use in the accessibility list.. essentially populating a `loadedGroups` object in the map state as chunks are loaded in with compressed data (id, lat/lng, filter). In future we could probably just tap into the apollo cache, or remove chunks as they are loaded in/ out of the map instead of letting this store grow and grow. Or something even better!

Aside that, we then know which toilets are loaded using the `loadedGroups` object. Then we can re-enable the `AccessibilityIntersection` component which cuts the `loadedGroups` down into 6 loos as before.

Following this the toilet names are fetched using a new GQL query: `useLooNamesByIds`.. the query accepts the 6 toilet IDs and gives us 6 names (hopefully) in return. This happens in `AccessibilityList` and is necessary because toilet names aren't available to us anymore due to our efforts to decrease the size of the payload sent to users.

** Side thought:** One possible optimisation here would be to have an option to load names/extra metadata selectively for some geohash chunks at source in `Markers.tsx`.. then we could load extra metadata exclusively for the chunk in the middle of the screen for example. I'm imagining something like:

```
  const { data } = useLoosByGeohashQuery({
    variables: { geohash, extraFields: "name, creationDate" },
  });
```

Where `extraFields` is optional extra stuff you'd like chucked over in our compressed loo format. Having the name fetch as a separate request  does work for now though!

I've tested it locally with VoiceOver on my MacBook.. it seems to work well. Although @falconx  you might like to check yourself!

Some screenshots:
Screenshot 2022-02-13 at 01 43 41
Screenshot 2022-02-13 at 01 44 15

Resolves #1273

### DIFF
--- a/src/api/schema.graphql
+++ b/src/api/schema.graphql
@@ -203,6 +203,7 @@ type Query {
     "A Point from which to begin the search"
     from: ProximityInput!
   ): [Loo!]!
+  looNamesByIds(idList: [ID]): [Loo!]!
   ukLooMarkers: [String!]! @cacheControl(maxAge: 360, scope: PUBLIC)
   "Retrieve a list of areas in existance, name and type only"
   areas: [AdminGeo!]!
--- a/src/api-client/operations/findLooNamesByIds.graphql
+++ b/src/api-client/operations/findLooNamesByIds.graphql
@@ -1,0 +1,6 @@
+query looNamesByIds($idList: [ID]) {
+  looNamesByIds(idList: $idList) {
+    id
+    name
+  }
+}
--- a/src/api/resolvers.ts
+++ b/src/api/resolvers.ts
@@ -1,5 +1,6 @@
 import { stringifyAndCompressLoos } from '../lib/loo';
 import ngeohash from 'ngeohash';
+import { async } from 'hasha';
 
 const { Loo, Report, Area, MapGeo } = require('./db');
 const { GraphQLDateTime } = require('graphql-iso-date');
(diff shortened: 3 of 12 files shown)


falconx (approved):
Looks good to me! The only thing I noticed was the list briefly displaying "Name Unknown" for each toilet. I guess ideally the list would be empty or show some loading state instead. That or your _side thought_, which I guess would avoid this. This is mostly insignificant though imo and less important than the screen reader announcements.

ob6160:
> Looks good to me! The only thing I noticed was the list briefly displaying "Name Unknown" for each toilet. I guess ideally the list would be empty or show some loading state instead. That or your _side thought_, which I guess would avoid this. This is mostly insignificant though imo and less important than the screen reader announcements.

thanks for the review!

that's a good point about the list, I'll change it so that it doesn't display until we have some names through before merging.

 will make a new ticket for tidying up the implementation here

src/api/schema.graphql
```diff
@@ -203,6 +203,7 @@ type Query {
     "A Point from which to begin the search"
     from: ProximityInput!
   ): [Loo!]!
+  looNamesByIds(idList: [ID]): [Loo!]!
```
marxian, line 206: A quibble... But isn't this more properly named loosByIDs ?

marxian (approved):

PR #1347, opened 2022-06-02
feat: configure sentry

marxian:
I think this'll establish some error monitoring and performance tracking.

Once it is in and we can see some output I might wish to get some release versioning going as well...

### DIFF
--- a/sentry.properties
+++ b/sentry.properties
@@ -1,4 +1,3 @@
 defaults.url=https://sentry.io/
 defaults.org=public-convenience-ltd
 defaults.project=toilet-map
-cli.executable=../../../.npm/_npx/a8388072043b4cbc/node_modules/@sentry/cli/bin/sentry-cli


sentry.properties
```diff
@@ -0,0 +1,4 @@
+defaults.url=https://sentry.io/
+defaults.org=public-convenience-ltd
+defaults.project=toilet-map
+cli.executable=../../../.npm/_npx/a8388072043b4cbc/node_modules/@sentry/cli/bin/sentry-cli
```
ob6160, line 4: this seems like a super specific path for npx — will this be stable, or should it hit just `node_modules/@sentry/cli/bin/sentry-cli`?

sentry.server.config.js
```diff
@@ -0,0 +1,21 @@
...
+import * as Sentry from '@sentry/nextjs';
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN;
+
+Sentry.init({
+  dsn:
+    SENTRY_DSN ||
+    'https://6935d02f879c4536aecfe8fa07ae1d3a@o1270901.ingest.sentry.io/6462188',
+  // Adjust this value in production, or use tracesSampler for greater control
+  tracesSampleRate: 0.1,
```
ob6160, line 14: do we only want to sample 10% of users?

marxian: We can't afford more on our free plan. Even this might be too much.

sentry.properties
marxian: The sentry tool generated this. I don't _think_ we even need it...

ob6160 (approved):
Lgtm 🙂

PR #1356, opened 2022-06-18
Transpile react-leaflet and markercluster 

ob6160:
## What does this change?

These modules were including the null coalesce operator in the build, meaning that the app crashed at runtime on older browsers. 

This change ensures that these modules are transpiled as part of our build step using `next-transpile-modules` https://github.com/martpie/next-transpile-modules

Tested in Chrome 72 and the website now renders the map instead of erroring at runtime


marxian (approved):
Oh well done!

I'd forgotten about next-transpire-modules!

PR #1361, opened 2022-06-19
Prevent re-render of map on route change

ob6160:
## What does this change?
We were re-rendering the entire leaflet map when changing route from `/` to `/loos/[id]`. 

This caused a noticeable flicker when clicking a marker for the first time.

This change lifts the map higher and passes it down just to the `/` and `/loos/[id]` pages so it isn't re-rendered when the page-props change when a new route is pushed (i.e onClick of a marker).

### Before

Uploading Screen Recording 2022-06-19 at 19.54.58.mov…

Uploading Screen Recording 2022-06-19 at 19.54.33.mov…

## How was this tested?
I've tested this change locally and through a preview deployment.


marxian (approved):
Neat. 

I think this is likely to get even smoother in future next.js versions 

https://nextjs.org/blog/layouts-rfc?utm_source=next-site&utm_medium=banner&utm_campaign=next-website

ob6160:
> Neat.
> 
> I think this is likely to get even smoother in future next.js versions
> 
> https://nextjs.org/blog/layouts-rfc?utm_source=next-site&utm_medium=banner&utm_campaign=next-website

ah! that's excellent, nice solution to the problem

PR #1373, opened 2022-07-08
Single mongoose connection

ob6160:
## What does this change?

I've changed our connection strategy for the main app so that we only open one mongoose connection to the mongo server for the lifetime of the application instead of per-request to `/api`. This will hopefully reduce the number of concurrent connections that we make.

## How was this tested?

I've tested this locally and on a staging instance.


ob6160:
I'm merging this to test it with real load. If things go wrong I will revert!

PR #1376, opened 2022-07-09
Cache geohash tiles and invalidate on add/edit/removal

ob6160:
## What does this change?

With these changes we re-enable caching on the server side for geohash queries.

These queries define the tiles that we draw to the map and are a core driver of connections made to mongo.

By re-enabling the caching layer in the gql schema for these we will hopefully drive down the number of requests we need to make to mongo for standard users who are just browsing for a loo.

## New stuff

This PR also adds the option to run the toilet map backed with a Redis store cache attached to the apollo server. This is beneficial because—by centralising our cache—we avoid having to make as many lookups for map tile loo information against our Mongo database which is currently expensive due to number of connections we incur.

So, with this change, if something is in cache (memory or Redis), we now *only* run our `dbConnect` method at the resolver level. This is key (if you'll excuse the pun) because that means that if there is a cache hit, we never need to connect to mongo to serve the request, we need only ask the in-memory cache or Redis. (@marxian this is the new work since you last reviewed)

Redis servers are quite cheap and, with a good caching layer, we could avoid having to scale the Mongodb side as much should traffic increase.

### Cache invalidation

We'd like to invalidate the cache for loo tiles when loos are added, updated or removed.

We do this at two levels by passing a new header: `invalidatecache` as part of the `useLoosByGeohashQuery` that we execute for each tile of markers requested by the client when a user pans/zooms/loads the map.

1. *HTTP Layer* :: If we're coming from an add/edit/remove page and the `invalidatecache` header is set, we include the following in the headers that we send to the api: `{ 'cache-control': 'no-cache', pragma: 'no-cache' }`. This will ensure that the http caching layer is bypassed. This happens here: https://github.com/neontribe/gbptm/pull/1376/files#diff-a1c81509bdeab83584c5c0d145ef8e1d499670dba35cfa988161e7c47cb46b3cR34
2. *Server cache layer* :: If we're coming from an add/edit/remove page and the `invalidatecache` header is set, we stipulate that we should not read from cache on the api layer through the `shouldReadFromCache` method that is part of the plugin: `responseCachePlugin`. This happens here: https://github.com/neontribe/gbptm/pull/1376/files

## What else?

Some bug fixes and general improvements are also included.

## How was this tested?

Locally and on a staging instance.

### DIFF
--- a/src/api/resolvers.ts
+++ b/src/api/resolvers.ts
@@ -108,7 +108,6 @@ const resolvers = {
     loosByProximity: async (_parent, args) =>
       await DBLoo.findNear(args.from.lng, args.from.lat, args.from.maxDistance),
     loosByGeohash: async (_parent, args) => {
-      console.log('resolution for: ' + args.geohash);
       const geohash: string = args.geohash ?? '';
       const current = ngeohash.decode_bbox(geohash);
 
--- a/src/api/resolvers.ts
+++ b/src/api/resolvers.ts
@@ -2,7 +2,13 @@
 import { stringifyAndCompressLoos } from '../lib/loo';
 import ngeohash from 'ngeohash';
 
-import { Loo as DBLoo, Report as DBReport, Area, MapGeo } from './db';
+import {
+  Loo as DBLoo,
+  Report as DBReport,
+  Area,
+  MapGeo,
+  dbConnect,
+} from './db';
 import { GraphQLDateTime } from 'graphql-iso-date';
 import without from 'lodash/without';
 import OpeningTimesScalar from './OpeningTimesScalar';
@@ -36,8 +42,13 @@ const looInfoResolver = (property) => {
 
 const resolvers = {
   Query: {
-    loo: async (_parent, args) => await DBLoo.findById(args.id),
+    loo: async (_parent, args) => {
+      await dbConnect();
+      return await DBLoo.findById(args.id);
+    },
     loos: async (_parent, args, context) => {
+      await dbConnect();
+
       const REQUIRED_PERMISSION = 'VIEW_CONTRIBUTOR_INFO';
 
       // Construct the search query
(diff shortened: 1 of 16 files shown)


ob6160:
HM, not working like it did locally — investigating

ob6160:
> HM, not working like it did locally — investigating

seems as though it's fixed following some changes

marxian (approved):
This looks like it'd work!

It seems a bit tricky to coordinate the timing of the cache-bypasses - but I struggle to think of anything neater :-)

src/api/resolvers.ts
```diff
@@ -36,8 +42,13 @@ const looInfoResolver = (property) => {
 
 const resolvers = {
   Query: {
-    loo: async (_parent, args) => await DBLoo.findById(args.id),
+    loo: async (_parent, args) => {
+      await dbConnect();
```
ob6160, line 46: here is where we shift the `dbConnect` call down into the resolver, so it isn't called until we've checked whether we have a hit in the cache store

src/components/LooMap/Markers.tsx
```diff
@@ -102,13 +102,33 @@ const MarkerGroup: React.FC<{
...
   const router = useRouter();
   const [mapState, setMapState] = useMapState();
+  const [lastContribution, setLastContribution] = useState<
+    number | undefined
+  >();
   const map = useMap();
 
   const { appliedFilters: filters } = mapState;
 
+  useEffect(() => {
```
ob6160, line 112: we use this information to detemine how long it's been since the last contribution the user has made. If it hasn't been too long since then, we want to pass `invalidateCache: true` so that their updates are reflected on everybodys' map.

src/pages/api/index.page.ts
```diff
@@ -61,10 +81,31 @@ export const server = new ApolloServer({
...
       user,
+      invalidateCache,
     };
   },
   introspection: true,
-  plugins: [responseCachePlugin({})],
+  plugins: [
+    responseCachePlugin({
+      generateCacheKey: ({ request, queryHash }) => {
+        const geohash = request.variables?.geohash;
```
ob6160, line 91: we generate a nicely named key based on the geohash and the staging environment

src/pages/api/index.page.ts
```diff
@@ -61,10 +81,31 @@ export const server = new ApolloServer({
...
+        if (geohash) {
+          return cacheKeyFromGeohash(geohash);
+        }
+        return queryHash;
+      },
+      shouldReadFromCache: async ({ context, cache, request }) => {
+        if (!!context?.user && context?.invalidateCache) {
```
ob6160, line 98: then, for most users who are not logged in we always say that they should read from cache.

the exception to this rule are if a user is logged in *and* they have recently set the `LAST_CONTRIBUTION` cookie, meaning that the `invalidateCache` context value is `true`

if this is the case, we remove the implicated tile from the cache and return `false`, meaning that Apollo Server will call the resolver anew to fetch the fresh information

ob6160:
closing this because I'm not happy with the broader implementation here and need to take a step back to rethink

PR #1401, opened 2022-08-21
Dynamic map bounds

majewskim:
## What does this change?

Map now dynamically zooms out to fit the bounds of the user's location neighbours

## How was this tested?

Manual testing

## Accessibility

## Documentation


src/lib/loo.ts
```diff
@@ -1,3 +1,4 @@
+import { LatLngLiteral } from 'leaflet';
```
ob6160, line 1:
```suggestion
import type { LatLngLiteral } from 'leaflet';
```

src/components/LooMap/useLocateMapControl.ts
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import L, { LatLngExpression, Map } from 'leaflet';
```
ob6160, line 2:
```suggestion
import L, { LatLngLiteral, Map } from 'leaflet';
```

I don't think we were using any of the other types provided in the `LatLngExpression`, so i'm putting this (there are two successive changes that i can't put as gh code suggestions) in to narrow it down to just the `LatLngLiteral` type that you used in the `fitMapBoundsToUserLocationNeighbours` method signature so we can satisfy the type checker.

ob6160: worth mentioning that the Typescript migration is still a WIP, but we're trying to keep any new code typesafe so we reduce tech debt as we slowly squash all our type errors

src/lib/loo.ts
```diff
@@ -64,3 +65,26 @@ export const filterCompressedLooByAppliedFilters = (
 
   return passesAll;
 };
+
+export const fitMapBoundsToUserLocationNeighbours = (
+  userLocation: LatLngLiteral,
+  map: L.Map
+) => {
+  //find neighbours of the user's location and set the map bounds to fit them
```
ob6160, line 73: similar comment to the one above about mentioning that the neighbours are map tiles

ob6160 (commented):
nice one, thanks so much for the contribution here — will definitely make peoples' experience better. a few minor comments, and then it should be good to go!

ob6160 (commented):
oh — one other thing, this might break some of the end to end Cypress tests around the geolocation functionality. let me know if you want any pointers getting a end to end test environment to work locally

majewskim:
> oh — one other thing, this might break some of the end to end Cypress tests around the geolocation functionality. let me know if you want any pointers getting a end to end test environment to work locally

Oh yeah that would be helpful. I tried running cypress open but it doesn't recognise the cypress.config.ts and is always trying to create new ones :/

ob6160:
> > oh — one other thing, this might break some of the end to end Cypress tests around the geolocation functionality. let me know if you want any pointers getting a end to end test environment to work locally
> 
> Oh yeah that would be helpful. I tried running cypress open but it doesn't recognise the cypress.config.ts and is always trying to create new ones :/

i'd be happy to have a call with you to get you set up :) 

sorry for all the update noise on the PR, it's the first time we've had an external contribution from a fork since setting up our new workflow, so have been changing a few things to get the GitHub Action workflows to run as expected

src/components/LooMap/LooMap.tsx
```diff
@@ -313,7 +313,7 @@ const LooMap: React.FC<LooMapProps> = ({
           }
         >
           {alwaysShowGeolocateButton && <LocateMapControl />}
-          <Media greaterThan="md">
+          <Media greaterThanOrEqual="sm">
```
ob6160, line 316: Everything is looking good with regards to the zoom-out on geolocation feature.

These controls were never designed to be shown on mobile - here’s an example of what they look like on mine to illustrate:

4CAE97B7-E3EE-49FE-AFD0-823C5F26C2D7
4904A53E-E13D-4040-967E-87EE1484DB06

normally they look alright, but we shift them up on desktop away from the toilet description when it’s opened which - as shown by the screenshot - looks a bit off on mobile.

i suggest we leave this at `md` so it’s hidden on mobile still for now; then we can open an issue or another pr to investigate ways we can include the controls in a nicer way… or make a decision as to whether we’d like to include them at all.

ob6160 (commented):
Just the change mentioned in the comment really, otherwise I’m happy for this to go in :)

ob6160 (approved):
looks good :) thanks so much for the contribution! 

I need to play with the github actions a bit more to properly run CI jobs for contributions from forked repos.. please bear with

PR #1464, opened 2023-03-06
Adding the *beginnings* of a new design system

rosejbon:
## What does this change?

This introduces the first move in the toilet map moving away from using Emotion to style the site. See this issue - https://github.com/public-convenience-ltd/toiletmap/issues/1446

It adds one "atom" component that is a `Badge` and adds css variables to our Global CSS file (which is currently being used by Emotion, but eventually we will want as it's own css file).

The suggestion is that we will move away from our Component library in a piecemeal approach using the [Atomic Design Methodology](https://atomicdesign.bradfrost.com/chapter-2/#the-atomic-design-methodology) and start with the smallest atomic bits of design and work up to the larger pieces as we go. We are currently experimenting with just using css and seeing how we go.

## How was this tested?

At the atomic stage, @ob6160 and I discussed that there wasn't a need to add tests, and that the use of Storybook will capture visual and layout regressions of these elements.

## Accessibility

The darkest colour on the pink/red background doesn't pass colour contrast, so I've suggested we use the pink/red outline with white background and dark blue text.

## Documentation

- [ ] I'm not sure whether we want to update documentation to make this change clear or maybe just generate more issues so folk can see the direction we are heading.
- [x] Added the component to Storybook


src/globalStyles.tsx
```diff
@@ -1,13 +1,23 @@
...
+//To do: when we have moved away from Emotion, think about what we do with the file
 
 // loosely based on https://hankchizljaw.com/wrote/a-modern-css-reset + a height declaration on the next container...
 const globalStyles = (
   <Global
     styles={css`
       @import url('https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;700&display=swap');
 
+      :root {
+        --blue-color: #0a165e;
```
marxian, line 12: I only have one concern in all of this excellent goodness.

marxian: Would there still be an advantage in referring to pallette variables by names which indicate their use  rather than their color?

e.g. "--primary-color" "--active-color" "--warning-color" "--success-color" ?

ob6160: agree with the above - if you'd like a reference point, here's how we approach this in Source:
- Palette and colour definition file: https://github.com/guardian/csnx/blob/main/libs/%40guardian/source-foundations/src/colour/palette.tsx#L384
- Storybook showing the palette: https://guardian.github.io/csnx/?path=/docs/source-foundations_palette--palette

The key difference is that there's a separation between the "palette" of available colours in the design system that we dip into to create a mapping. https://github.com/guardian/csnx/blob/main/libs/%40guardian/source-foundations/src/colour/palette.tsx#L4 

For our use here though, I'd be a proponent of skipping the "palette" data structure and just referencing the hex codes directly, even if there is repetition

rosejbon: I know what you mean, I was thinking we could document them in Storybook under this is how we use our colours instead. I've found that in projects I end up having a random addition of colour variables anyway so it felt nice to have them all colours. 

On the flip side, if we ever wanted to update the styles of the site it is probably nicer if we had variable names that indicate use. I could ask Rachel if she had a list of palette variables she'd expect to reach for.

rosejbon: I'm not particularly wedded to either way, happy to go with the majority on this.

src/pages/_app.page.tsx
```diff
@@ -8,6 +8,7 @@ import '@fortawesome/fontawesome-svg-core/styles.css';
 config.autoAddCss = false; // Tell Font Awesome to skip adding the CSS automatically since it's being imported above
 import { useRouter } from 'next/router';
 import LooMap from '../components/LooMap/LooMapLoader';
+import '../design-system/components/Badge/Badge.css';
```
ob6160, line 11: I think this style sheet will need to be included in the Badge story in Storybook for it to be styled correctly there (I've added comments to the two affected stories in the Chromatic visual diff tool as well)

I think something like this near the top in `.storybook/preview.js` will do the trick nicely:

```typescript
// Import design system component styles for our Stories.
import '../src/design-system/components/Badge/Badge.css';
```

We'll just need to make sure to keep both `_app.page.tsx` and `.storybook/preview.js` up to date with the latest component styles until we combine them all into a single design system stylesheet

src/design-system/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+  title: 'Badge',
+  component: Badge,
+} as ComponentMeta<typeof Badge>;
+
+const Template: ComponentStory<typeof Badge> = (args) => <Badge {...args} />;
+
+export const Primary = Template.bind({});
+
+Primary.args = {
+  text: 'Primary badge',
```
ob6160, line 15:
```suggestion
  children: 'Primary badge',
```

I think this got missed with the switch from `text` to accepting `children` 

Without it, the story won't render any text inside the badge

ob6160 (commented):
great work! this sets us nicely on the path towards our new design system 💎 

I left two comments around the Storybook configuration that we'll need to address so the story is rendered correctly.

I think the only other thing that stood out was @marxian 's suggestion to change the colour names

After those are resolved, I'm happy to ✅

src/design-system/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+
+import Badge from './Badge';
+
+export default {
+  title: 'Badge',
```
ob6160, line 6:
```suggestion
  title: 'Components/Badge',
```

If we change the title to include that prefix, I think it'll fit nicely under the existing components section.

Perhaps we should make a new section though, to keep new components separate from the Emotion ones?

src/design-system/components/Badge/Badge.css
```diff
@@ -0,0 +1,7 @@
+.badge {
+  padding: 4px;
```
ob6160, line 2: Looking forward hearing about your planned exploration around the css vars for spacing and sizes!

src/pages/_app.page.tsx
rosejbon: 👍🏻 good spot, I hadn't checked Storybook after I fixed the css conflict error I was having. 

I've managed to import the css into the actual story (`./Badge.stories.tsx`), would there be an advantage to putting the css there rather than in `.storybook/preview.js`,  in that folk might be more likely to remember to add it, or is that an odd place to have it?

ob6160: If it's only imported in `./Badge.stories.tsx` we risk it working for that story, but we have a story which uses the Badge component such as the toilet details panel, the styles might not be available because they're coupled tightly to the Badge story only, not the ToiletDetailsPanel story.

Having it in `.storybook/preview.js` ensures that the badge styles are available to every story that we define (kinda analogous to putting into `_app.page.tsx` in the context of the Next app)

rosejbon: 👍🏻 That makes total sense, thank Ollie 😄

src/design-system/components/Badge/Badge.stories.tsx
rosejbon: We could make a new section, I think that makes sense and then slowly move stuff from components away to it, would be a nice visual way to see the work done/to be done in Storybook, maybe something like `Design-System/Badge`?

ob6160: I think that sounds reasonable :)

src/globalStyles.tsx
rosejbon: Ok, so Rachel and I have been having a chat. It doesn't seem there is a general consensus to this. Quick summary of the pros and cons of the approaches:

### Variable names that indicate use
#### Strengths 
- We could be very consistent in how we use the colour palette 
- Allows us the ability to quickly change up the look and feel of the site. 
#### Drawbacks
- Systems designed this way can be quite rigid 
- There could be an amount of up front work suggest by Rachel in mapping out the different states we may want or if we add variables as we go along this could lead to inconsistencies depending on the preference of different developers/designers.

### Colour variable names
#### Advantages
- The fairly small colour palette we have means a relatively small set of variables to use
- Allows designers and developers to combine colours in whatever way they wish without adding new variable names
#### Disadvantages
- Use of written documentation to indicate how they should be used
- Potential for colour combinations to be combined in an inaccessible way

I think after chatting with Rachel that colour variable names feel more flexible for an open source project which may have a variety of different people inputting into it, especially as it is a relatively small site with a fairly modest colour palette. I'd be interested in experimenting with just colour variables and seeing if that causes confusion or doesn't work when we grow the design system.

ob6160: Agreed, happy to run with the colour variable names for now - let's see how it plays out :)

ob6160 (approved):
🎨

PR #1468, opened 2023-03-10
Remove references to mongodb

ob6160:
## What does this change?

Now that we're fairly confident the migration has gone well this is a general tidyup mongodb related scripts, dependencies and documentation. 

This tag: https://github.com/public-convenience-ltd/toiletmap/releases/tag/mongo-checkpoint captures today's version of the loo map - which has a working version of the migration script, should we need to run it again in the future.

Summary:

- Depends on https://github.com/public-convenience-ltd/toiletmap/pull/1467
- Removes the migration script from mongodb to postgres
- Swaps references from mongodb to postgres in our documentation
- Provides a new setup guide that lists Docker as a pre-requisite for the local dev env
- Uninstalls mongoose et al.


marxian (approved):
I'd say you got it all!

Issue #1463, opened 2023-03-05
Allow users to zoom out further in the edit/add toilet view


ob6160:
- The map on the add/edit toilet pages is limited to only a rather close-in zoom level. This makes it difficult to get your bearings when trying to find the exact location of the toilet you're looking to add
- Let's look at allowing users to zoom out a little further on this page?

PR #1483, opened 2023-03-30
fix: add/edit page min-zoom consistent with index page

Ocoldwell:
## What does this change?

This should solve #1463 and is an adjustment to the editMinZoom in config.ts which makes the min zoom consistent across index and add/edit page.

## How was this tested?

Tested in local dev environment as change is minor.

## Accessibility

## Documentation


ob6160 (approved):
Thanks for taking the time, looks good to me 💎

Issue #1453, opened 2023-03-04
Complete NextJS 13 migration


ob6160:
- At some point I upgraded NextJS and I was lazy and didn't migrate image imports: https://github.com/public-convenience-ltd/toiletmap/blob/main/src/components/Footer.tsx#L6 
- We also use `legacyBehaviour` on links: https://github.com/public-convenience-ltd/toiletmap/blob/70568dc8e3bdbf93ca3eaa243a9a2de6058286bd/src/pages/login.page.tsx#L46
- I probably missed some other things.. need to put some time into seeing what
- Migration guide: https://beta.nextjs.org/docs/upgrade-guide#image-component

PR #1484, opened 2023-04-02
Nextjs 13 migration - Image and Link updates

Ocoldwell:
## What does this change?
These changes mostly refer to the issue #1453 and there are three main refactors in this PR: 
- We are updating the Image component to use the next/image import and the non-legacy api, this deprecates the layout property and we need to apply the styles from this in either a style or as css. I ran the codemod on the migration guide which suggested the styles in Logo.tsx. 
- The second change is updating the Link behaviour so that we no longer are using legacyBehaviour.  We no longer need to use the passHref property or an `<a>` tag as a child. I've tried to keep the DOM structure as simple as possible, we can use the Button component to pass the Link which allows us to keep the styles of the button variants and ensuring we aren't including `<button>` tags inside of `<a>` tags.
- The third change is about consistency, and can be changed quickly if preferred, but I found that the email links did not have consistent behaviour so I've updated all email links to have the same behaviour

## How was this tested?
Local testing and checking in dev tools as I'm currently unable to run the Cypress e2e tests for any regressions but I've tried to ensure the same DOM structure and simplified where I can.

## Accessibility
Changes all resolve to valid <a> tags with no nested buttons inside.

## Documentation

### DIFF
--- a/src/components/EntryForm.tsx
+++ b/src/components/EntryForm.tsx
@@ -6,7 +6,7 @@ import {
   UseFormRegister,
   FieldValues,
 } from 'react-hook-form';
-import Image from 'next/legacy/image';
+import Image from 'next/image';
 import isFunction from 'lodash/isFunction';
 import omit from 'lodash/omit';
 import pick from 'lodash/pick';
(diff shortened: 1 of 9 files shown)


ob6160:
heya, thanks so much for taking the time here! ✨ I'll get a proper review in soon!

I'll look also look into getting authorised GH workflow runs working so they can access the secrets they need to run the Cypress tests. We don't often get contributions so - as you can probably tell - the workflow isn't tried and tested here. Please bear with me :)

Issue #1462, opened 2023-03-05
Improve the opening times input UX


ob6160:
## Background 
Right now toilet opening times have to be painstakingly input one-by-one, leading to it becoming a fairly time consuming process.

A recent piece of anonymous user feedback further highlights the need for some improvements here:

> When editing opening hours can we have a way to say its the same every day to save typing it out multiple times? I see some people are putting opening in the comments and this may be why.

Here's an example of what the user interface looks like currently:
<img width="874" alt="Screenshot 2023-03-05 at 16 04 09" src="https://user-images.githubusercontent.com/1771189/222971691-9ae1c491-7612-4033-864e-aeaed3f5b228.png">

It takes fourteen discreet actions to input each of the open/close times for every day in the week.

How can we improve this experience for our users? Can we perhaps allow for a single range (0700—1700) to be repeated across each of the input fields?

Ocoldwell:
I've a few approaches that we could try and they might help with the user experience:

- When someone knows the opening times, it might be slightly better UX for the opening times to be open by default, as that's up to an additional 7 actions a user must make to enter data for a week. This may create data issues of it's own, wherein a user manually inputs the times but doesn't remember to close a day of the week.

- We could have an additional input which sets a global/default range at the top of the form, when this is updated it will fill all time range inputs where a toilet is open with this value,  which then allows a user to then fine-edit as necessary.

- Another option we could try is for the user to set the first range themselves, and then when they select the next row of inputs the previous value will populate the row range which they then can adjust as necessary.

- We could also try having a toggle switch with the question "Is it open the same time each day?" which the user can click which would reduce the inputs to one which they can then edit, behind the scenes that populates the form week days with the opening time.

ob6160:
Thanks both, @harryharrold good shout. @Ocoldwell thanks for the considered list of ideas.

When I get a mo I'll run some queries against the db so we can determine what people usually put down to better evaluate our options.

Ocoldwell:
Another idea we could try could include an edit all tickbox/toggle which when toggled on will mean any changes to time inputs will edit all time inputs for the week. Then if fine grain changes are needed they can toggle it off.

ob6160:
[...]

PR #1487, opened 2023-04-06
feature: Added an edit all checkbox for opening times form

Ocoldwell:
## What does this change?

This introduces the edit all checkbox discussed in #1462, which works by adding a function when the checkbox is ticked that will iterate over all open toilets inputs of the types -opens or -closes and edits them with the inputted value. This also introduces DX changes by introducing a defaultFormState to the useForm (this gives something for dirtyFields to be compared against) and types which will help with further refactoring.

Further styling changes so that the weekdays are padded and nested to be in line with the rest of the form and some small performance improvements by reducing the number of string.prototype.toLowerCase() calls in existing functionality.

## How was this tested?
Local testing and an untested e2e test for deployment testing.

## Accessibility
Using native html checkbox and navigable by keyboard.

## Documentation
No changes at the moment.


ob6160 (approved):
great work, thank you for the contribution :)

PR #1491, opened 2023-04-27
Add button to new design system

rosejbon:
## What does this change?

Adds a new Button to the design system - this button has two style variants like the existing button; `primary` and `secondary`. It also can be used a `button` or a `link`. 

I have removed the `link` style variant, and where a `Button` was being used as a `link` and styled like I `link`, I've just used the [NextJS Link](https://nextjs.org/docs/api-reference/next/link) component instead, which is styled to appropriately by our global css.

## How was this tested?

- Hoping our cypress tests will pick anything egregious 
- But also using Chromatic and testing this update on a preview deployment

## Accessibility

## Documentation


ob6160 (commented):
great work! There was just one page (edit I think) which I've commented on in Chromatic where the two buttons look a bit off. Otherwise looking good to go :)

ob6160 (approved):
lovely!

PR #1539, opened 2023-08-17
style: remove max-width for buttons

rosejbon:
If we want specific sizes for buttons we should add new variants. I think for now we shouldn't constrain the size of the buttons rather let them either fill the space given by the parent or be as large as they need to be as determined by their children.

## What does this change?

Removes the max-width for buttons

## How was this tested?

- Testing via Chromatic as this is only a visual change to the width of the button

## Accessibility

n/a

## Documentation

n/a


ob6160 (approved):
🫡

Issue #1543, opened 2023-09-13
Feat: add short text input 


rosejbon:
Part of the [move away from Emotion to set up our design system](https://github.com/public-convenience-ltd/toiletmap/issues/1446). This is the work to add a short text input to the design system styling it with just css and to replace existing short text inputs with this new component.  See the [Badge](https://github.com/public-convenience-ltd/toiletmap/tree/main/src/design-system/components/Badge) component as an example. 

### Design
![image](https://github.com/public-convenience-ltd/toiletmap/assets/44296500/4a47635e-319f-4d80-8238-9cbe63881b73)

PR #1547, opened 2023-09-14
Refactor input field styles

hannahouazzane:
## What does this change?

#1543 
Added input field to design system using CSS styling only. 
Replaced existing input field with new component

## How was this tested?

Tested change on a preview deployment

## Accessibility

Checked to see if navigable by keyboard both manually and by using dev tools
Colour contrast is accessible according to: https://colourcontrast.cc/ffffff/0a165e

## Documentation
 Added input field to Storybook


ob6160 (approved):
looks good! :shipit:

PR #1548, opened 2023-09-19
Add icons to new design system

rosejbon:
## What does this change?

Adds a new Icon component to the design system and removes the need to use Font Awesome packages. We were only using a small number of icons and can add ones when we need them. There are some slight variance in the icons used and the size is slightly different. This component has the following props:

- size
- icon
- spin: a boolean for whether the animation should spin or not

It inherits the colour of the parent it is in. 

## How was this tested?

- Have tested with Cypress (having issues with that)
- Chromatic

## Accessibility

When swapped out I have made sure that any icons have `aria-labels` if needed or even better are accompanied with visual text (or hidden text for screen readers). This component will require future devs to correctly use it to make sure it is accessible. 

## Documentation

- Added to Storybook


ob6160 (commented):
looks good! just a few things slightly off in the Chromatic UI diff

hannahouazzane (approved):

ob6160 (approved):
great work!! :) :shipit:

PR #1551, opened 2023-09-20
Potential fix for failing edit test on firefox

ob6160:
## What does this change?

- Catch spurious errors in our Nominatim location fetch routine which was causing our Firefox Cypress test to fail on the uncaught exception


rosejbon (approved):
🥳

PR #1557, opened 2023-09-21
Remove pr_target privileged run label check for Cypress Actions

ob6160:
## What does this change?

- This was adding a lot of complexity for a use case that doesn't happen very often.
- Pass `--headed=true` to fix an issue with Cypress v11 when using Chrome 117 (issue related: https://github.com/cypress-io/cypress/issues/27804#issuecomment-1721476731)


hannahouazzane (approved):
🥳 🙌 woohoo!

PR #1573, opened 2023-10-03
Package updates for eslint and prettier packages. And minor package updates.

rosejbon:
## What does this change?

Package updates for eslint and prettier packages. And minor package updates.

## How was this tested?

- End to end cypress tests (It is passing on all Chrome Cypress tests, but failing on all Firefox Tests, think this is a part of work to improve Cypress tests)
- Locally


ob6160 (approved):
lovely :) 
Chrome cypress tests good enough for me!

PR #1653, opened 2024-04-18
Add drawer component to design system and site

rosejbon:
## What does this change?

Continues the work of updating the design system. Adds the existing component.

#### Things to note

Things to note, I decided to basically copy the existing component over. I think there are accessibility issues with how it is currently used in the hamburger nav and filter on mobile, mainly as it covers the screen but elements remain navigable under it. I have two thoughts on this:

1. We could refactor the header and filter on mobile to use a focus trap to make sure users can only tab on what they see, and also make sure that the key board nav aligns with accessibility standards.
2. We wait for the [PopOver api](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) to be fully supported which will fix all these accessibility issues.

I think realistically we need to do 1 before we refactor with the second option.

## How was this tested?

- Running Cypress end to end tests and passed locally
- Checking on my machine that is looks good

## Accessibility

See above comment

## Documentation

Component has been added to Storybook


src/design-system/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,39 @@
+import DrawerProps from './Drawer.types';
+
+import { motion } from 'framer-motion';
```
ob6160, line 3: hmmm, it goes a bit beyond the scope of the work here, but maybe we could remove `framer-motion` as a dependency?

I _thiink_ it's only used here and it might reduce our bundle size a bit + remove a bit of an overkill library that just animates the drawer coming in and out.. something we could probably do quite easily with CSS transitions perhaps? :)

ob6160 (approved):
nice one, thanks for continuing with the great work on this :) one small comment / suggestion, otherwise looks good to go!

src/design-system/components/Drawer/Drawer.tsx
rosejbon: Yeah, I had wanted to strip it out, but struggled to get the css animation to do what I want. Think I've had an idea for how we might approach it. I'll merge it in and spin up a separate PR for that. 😄

ob6160: sounds good! :)

PR #1663, opened 2024-05-10
Use protomaps for the tiles layer

mxdvl:
## What does this change?

Use the [protomaps](https://protomaps.com/) tile base layer instead of OSM:

- free non-commercial API usage
- [“light” theme](https://maps.protomaps.com/#map=14.32/51.50237/-0.13745&theme=light) has less information overload
- support for high-DPI / retina displays out of the box
- we can customise map styles – #1670 

| Before | After |
|--------|--------|
| osm | protomaps | 

## How was this tested?

How have you tested your changes?

- `pnpm dev`
- local Storybook

## Accessibility

- [X] Ensured that the colour contrast is passing

## Documentation

- [X] Added/edited the appropriate documentation
- [X] Added any component(s) to Storybook


marxian (changes requested):
Hey @mxdvl!

Thanks for this :-) It certainly looks lovely.

I have some worries about color contrast accessibility and also about the removal of map features like landmarks.
I fear that these two aspects will cause problems for folk using the map to find a loo and navigating to it.

Can we affect these aspects?

mxdvl:
> I have some worries about color contrast accessibility and also about the removal of map features like landmarks. I fear that these two aspects will cause problems for folk using the map to find a loo and navigating to it.
> 
> Can we affect these aspects?

It’s definitely possible to adjust.

I’ve created two alternative base maps with different colour schems:
- #1668 
- #1667 

For Points of Interests (POI) the data comes from OSM all the same, so I’ll have a [look at adding things like landmarks](https://docs.protomaps.com/pmtiles/leaflet#custom-vector-data)

ob6160 (dismissed):

ob6160 (commented):
Looking good!

I think it'd be good to persist the users' choice in local storage. It seems odd to me that I'd want to opt-in only to lose my opt-in on the next reload

ob6160 (approved):
I love this, can't wait to see how it goes!
🗺️ 💎

marxian (approved):

PR #1664, opened 2024-05-10
Use a custom cluster marker style

mxdvl:
## What does this change?

| Before | After |
|--------|--------|
| before | after |

## How was this tested?

How have you tested your changes?

- Locally with `pnpm dev`

## Accessibility

- [X] Ensured that the colour contrast is passing

## Documentation

### DIFF
--- a/src/components/LooMap/Markers.tsx
+++ b/src/components/LooMap/Markers.tsx
@@ -45,7 +45,7 @@ const Markers = () => {
       boundingBoxWest,
       boundingBoxNorth,
       boundingBoxEast,
-      geohashPrecision
+      geohashPrecision,
     );
   }, [
     boundingBoxEast,
@@ -131,7 +131,7 @@ const MarkerGroup: React.FC<{
           }),
           alt: 'Public Toilet',
           keyboard: false,
-        }
+        },
       )
         .on('click', () => {
           // Clear the current search upon navigation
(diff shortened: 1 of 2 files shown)


marxian (approved):
This looks great!

Thanks @mxdvl

mxdvl:
I’ve accepted the Chromatic diffs as they were only changes to the OSM map updates – but there isn’t a story capturing the marker styles. Should I create one?

ob6160:
> I’ve accepted the Chromatic diffs as they were only changes to the OSM map updates – but there isn’t a story capturing the marker styles. Should I create one?

If it’s not too much extra effort to chuck in an extra map story with a marker cluster on it there’s no harm IMO. 

 I’m not massively bothered though as it should be quite obvious very quickly when testing if clusters look off if we capture it in Chromatic or not as it’s so core to the experience. Happy either way :)

mxdvl:
> If it’s not too much extra effort to chuck in an extra map story with a marker cluster on it there’s no harm IMO.

I’ve tried a few things, but I’m not understanding how to add marker groups to a story, so I’d prefer getting this in first and possibly improving this as a follow up.

ob6160:
> > If it’s not too much extra effort to chuck in an extra map story with a marker cluster on it there’s no harm IMO.
> 
> I’ve tried a few things, but I’m not understanding how to add marker groups to a story, so I’d prefer getting this in first and possibly improving this as a follow up.

sounds good - let's merge it!

👏 first loo map contribution 🚽

PR #1695, opened 2024-07-27
Add optional email field for feedback, record feedback in db

ob6160:
## What does this change?

- This improves the feedback form, adding an optional email field so we can get in touch with individuals who wish to be contacted back
  - This has been covered by changes to our privacy policy in: #1715  
- We also record the route so we know which page the user sent the feedback from
- Some small improvements to accessibility and width on mobile

Screenshot 2024-09-01 at 17 05 26

### DIFF
--- a/schema.prisma
+++ b/schema.prisma
@@ -89,8 +89,8 @@ model record_version {
 model feedback {
   id            Int       @id @default(autoincrement())
   email         String?   @db.VarChar(255)
-  feedback_text String
-  route         String
+  feedback_text String    @db.VarChar(5000)
+  route         String    @db.VarChar(255)
   created_at    DateTime? @default(now()) @db.Timestamp(6)
 
   @@schema("public")
--- a/schema.prisma
+++ b/schema.prisma
@@ -85,6 +85,17 @@ model record_version {
   @@schema("audit")
 }
 
+/// This model contains row level security and requires additional setup for migrations. Visit https://pris.ly/d/row-level-security for more info.
+model feedback {
+  id            Int       @id @default(autoincrement())
+  email         String?   @db.VarChar(255)
+  feedback_text String
+  route         String
+  created_at    DateTime? @default(now()) @db.Timestamp(6)
+
+  @@schema("public")
+}
+
 enum operation {
   INSERT
   UPDATE
--- a/src/pages/api/feedback/index.page.ts
+++ b/src/pages/api/feedback/index.page.ts
@@ -2,14 +2,19 @@ import { NextApiRequest, NextApiResponse } from 'next';
 
 async function handler(req: NextApiRequest, res: NextApiResponse) {
   const feedback = req.body;
-  fetch(process.env.SLACK_FEEDBACK_WEBHOOK, {
-    method: 'POST',
-    headers: {
-      'Content-Type': 'application/json',
-    },
-    body: JSON.stringify({ text: feedback?.text }),
-  });
-  res.send(200);
+  try {
+    await fetch(process.env.SLACK_FEEDBACK_WEBHOOK, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({ text: feedback?.text }),
+    });
+    res.send(200);
+  } catch (error) {
+    console.error('There was an error sending the feedback to Slack', error);
+    res.send(500);
+  }
 }
 
 export default handler;
--- a/src/pages/api/feedback/index.page.ts
+++ b/src/pages/api/feedback/index.page.ts
@@ -1,19 +1,43 @@
 import { NextApiRequest, NextApiResponse } from 'next';
+import prisma from '../../../api/prisma/prisma';
 
 async function handler(req: NextApiRequest, res: NextApiResponse) {
-  const feedback = req.body;
+  const { text, email, route } = req.body;
+
+  if (typeof text !== 'string' || text.length === 0) {
+    return res.status(400).send('Feedback text is required');
+  }
+
   try {
+    // We'd like to record the full feedback entry in our database for future reference.
+    await prisma.feedback.create({
+      data: {
+        email: email ?? 'No email provided',
+        feedback_text: text,
+        route,
+      },
+    });
+
+    if (!URL.canParse(process.env.SLACK_FEEDBACK_WEBHOOK)) {
+      console.warn('Slack feedback webhook not set, skipping send to channel');
+      return res.send(200);
+    }
+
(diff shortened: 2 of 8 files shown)


src/pages/api/feedback/index.page.ts
```diff
@@ -1,15 +1,44 @@
...
+    // We only pass on the feedback content and website path to Slack, not the users' email if they provided one.
+    await fetch(process.env.SLACK_FEEDBACK_WEBHOOK, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        text: `${text}\r\n------\r\nRoute: ${route}`,
+      }),
+    });
```
mxdvl, line 35: Do we want to `catch` here? Failing the webhook should not fail the submission…

mxdvl (approved):
db: the SQL

src/pages/api/feedback/index.page.ts
```diff
@@ -1,15 +1,44 @@
...
-    headers: {
-      'Content-Type': 'application/json',
-    },
-    body: JSON.stringify({ text: feedback?.text }),
-  });
-  res.send(200);
+  const { text, email, route } = req.body;
+
+  if (typeof text !== 'string' || text.length === 0) {
+    return res.status(400).send('Feedback text is required');
```
Hoolean, line 8: as we're adding the check, we might want to bubble this up to our form on the frontend

src/pages/api/feedback/index.page.ts
```diff
@@ -1,15 +1,44 @@
...
+  }
+
+  try {
+    // We'd like to record the full feedback entry in our database for future reference.
+    await prisma.feedback.create({
+      data: {
+        email: email ?? 'No email provided',
```
Hoolean, line 15: can we make this column nullable instead?

supabase/migrations/20240823182342_feedback_table.sql
```diff
@@ -0,0 +1,33 @@
...
+ALTER TABLE IF EXISTS public.feedback OWNER TO postgres;
+GRANT ALL ON TABLE public.feedback TO anon;
+GRANT ALL ON TABLE public.feedback TO authenticated;
+GRANT ALL ON TABLE public.feedback TO service_role;
+GRANT ALL ON TABLE public.feedback TO postgres;
+
+GRANT SELECT ON TABLE public.feedback TO toiletmap_web;
+GRANT INSERT ON TABLE public.feedback TO toiletmap_web;
+GRANT USAGE, SELECT ON SEQUENCE public.feedback_id_seq TO toiletmap_web;
+
```
Hoolean, line 20: not a review but a personal Q as I'm not so familiar with `GRANT`, what does limiting `toiletmap_web` to these exclude it from doing? :)

src/pages/api/feedback/index.page.ts
ob6160: agreed ✨

Hoolean (commented):
Nice :muscle: Added a few comments on validation and column types (and some bits I am just curious about); nothing blocking though, otherwise looks good to me

Hoolean:
Whoopsy, I've double-reviewed you - definitely not meant to replace yours Max, just had the page open too long before hitting go :smile:

ob6160:
> Whoopsy, I've double-reviewed you - definitely not meant to replace yours Max, just had the page open too long before hitting go 😄

thanks for the extra comments, all good points - i'll get back soon 🚗

src/pages/api/feedback/index.page.ts
ob6160: nice catch, this is nullable by default so I didn't really need to have this text fallback. if `email` is undefined/null this field will be null too

src/pages/api/feedback/index.page.ts
ob6160: I'll add a frontend check for text in the textarea before a submission is allowed to be submitted, so we have this backend check as well as a frontend message

ob6160: here's the check:

https://github.com/public-convenience-ltd/toiletmap/pull/1695/files#diff-c490ab266c5906e882ea09e434550470651318e8a7c87b02823b61ecb1fbcf18R63-R65

supabase/migrations/20240823182342_feedback_table.sql
ob6160: By default the role will not be given rights to SELECT/INSERT/UPDATE/DELETE/EXECUTE/etc... on any given table. We use the `GRANT` system to selectively allowlist our `toiletmap_web` role to perform only the operations it needs in order to function as intended.

In this case we'd like to allow grant it:
- Select & insert permissions
- Select & usage on the feedback id sequence

ob6160: Examples of what it's excluded from doing..

- No UPDATE or DELETE permissions.
- It can't modify the table schema because it lacks ALTER permissions.
- It can't run functions or procedures on that table.
- It can't drop things

src/pages/api/feedback/index.page.ts
ob6160: this is now implemented

schema.prisma
```diff
@@ -85,6 +85,17 @@ model record_version {
   @@schema("audit")
 }
 
+/// This model contains row level security and requires additional setup for migrations. Visit https://pris.ly/d/row-level-security for more info.
+model feedback {
+  id            Int       @id @default(autoincrement())
+  email         String?   @db.VarChar(255)
```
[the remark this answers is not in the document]
ob6160: I've added `maxlength` to appropriate inputs :)

schema.prisma
```diff
@@ -85,6 +85,17 @@ model record_version {
...
+  feedback_text String
+  route         String
```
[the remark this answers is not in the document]
ob6160: I've limited feedback_text to 5000 chars and route to 255, both seemed reasonable. If someone has more than 5000>chars to write they better write us an email

ob6160:
@Hoolean I've had a look and made some changes based on your feedback :)

src/pages/api/feedback/index.page.ts
Hoolean: Probably not blocking though, would be happy if we span out into a separate issue :)

Hoolean (approved):
All feedback addressed :shipit:

src/pages/api/feedback/index.page.ts
ob6160: sounds good, I'll make an issue to capture 🆒

PR #1706, opened 2024-08-17
ob/psql overhaul

ob6160:
## What does this change?

- **create cleaner inital db migration which combines the two previous**
- **add row level security and dev server certificates to gitignore**


charlesstrange2 (approved):
Ok, i think this should do what you want 👍

PR #1716, opened 2024-09-23
Various upgrades 

ob6160:
## What does this change?

- Bumps to Next 14
- Bumps various dependencies to their latst minor/patch versions
- Bumps node to the latest LTS version (20.9.0)
- Removes sentry package (and configuration) ... we are no longer using it


.nvmrc
```diff
@@ -1 +1 @@
-18.12.1
+20.9.0
```
mxdvl, line 1: Go bold: 
```suggestion
20.17.0
```

package.json
```diff
@@ -85,34 +84,33 @@
     "leaflet": "^1.9.4",
     "leaflet.markercluster": "^1.5.3",
     "lodash": "^4.17.21",
-    "monaco-editor": "^0.37.1",
-    "next": "^13.5.4",
-    "next-plausible": "^3.12.0",
+    "monaco-editor": "^0.52.0",
```
mxdvl, line 87: This could technically be breaking

mxdvl (approved):
I'd recommend [pinning dependencies in `package.json`](https://docs.renovatebot.com/dependency-pinning/). Also curious about the use of `devDependencies` — what do they represent?

.nvmrc
ob6160: agree let's go forth boldly 🖖 🫡

package.json
ob6160: good point — we just use this to render the nice json preview here: https://toiletmap-ey2ebn145-public-convenience-ltd.vercel.app/explorer/loos/fe9b7bb8acad4acee7eafcd8 and all seems well

ob6160:
> I'd recommend [pinning dependencies in `package.json`](https://docs.renovatebot.com/dependency-pinning/). Also curious about the use of `devDependencies` — what do they represent?

> pinning
yes I agree, has been on my to-do list for a while, I'll raise a followup when I have the time

> devDependencies
these have become a bit of a mess over time, some deps for prod are in the dev deps — another followup will be to sort through these and work out which are _truly_ dev deps

both followups open to others contributing, let me know if you fancy picking something up!

PR #1717, opened 2024-09-23
Remove firefox from our ci pipeline

ob6160:
## What does this change?

- We're doubling up our end to end tests for both Chrome and Firefox for arguably minimal gains. Firefox is a modern browser and if the tests pass against Chrome it's very likely the application will be working against Firefox too.
- It means I get to remove lines of code 😈 
- We're less likely to experience e2e test flake + we can remove some skipped tests on FF / setup quirks


marxian (approved):
Agree...

It is unlikely that this extra coverage provides extra confidence in our situation.

PR #1725, opened 2024-09-25
Use edge geolocation to center the map

marxian:


src/middleware.page.ts
```diff
@@ -0,0 +1,16 @@
...
+
+// run only on homepage
+export const config = {
+  matcher: '/',
+};
+
+export async function middleware(req: NextRequest) {
+  const { nextUrl: url, geo } = req;
+  console.log(geo);
+  if (!url.searchParams.has('lat') && geo && geo.latitude && geo.longitude) {
```
mxdvl, line 11: FYI – 0,0 is a valid lat/lng, so it’s probably best to check whether these are `undefined`.

Alternatively if you want absolute type-safety using `zod`, you could go down this route:

```ts
import z from 'zod';

const geoSchema = z.object({
  latitude: z.number();
  longitude: z.number();
})

const geoResult = schema.safeParse(geo);
if(geoResult.success && !url.searchParams.has('center')) {
  const { latitude, longitude } = geoResult.data;
  url.searchParams.set('center', [longitude, latitude].join(','));
}

const centerSchema = z.string()
  .transform((value) => value.split(",").map((l) => parseFloat(l)))
  .pipe(z.tuple([z.number().finite(), z.number().finite()]));
  
const centerResult = centerSchema.parse(router.query.center)
if(centerResult.success) {
  const [lng, lat] = centerResult.data;
  console.log({ lng, lat });
}
```

src/middleware.page.ts
```diff
@@ -0,0 +1,16 @@
...
+    url.searchParams.append('lat', geo.latitude);
+    url.searchParams.append('lng', geo.longitude);
```
ob6160, line 13: I'm not sure I agree with appending this to the `searchParams`.

A user who is on the go and using their mobile might reload the page at some point, but because lat/lng are already set as params they won't update to the latest edge location.

Also, people who share the link to the site with their friends will likely bring the lat/lng along if they copy/paste the link from their browser meaning that people who land on the site will get the wrong initial location for them

PR #1727, opened 2024-10-06
Set gh actions cypress action to v6 

ob6160:
## What does this change?

- We've been having persistent issues with our Cypress binary cache not being persisted properly across CI runs, hopefully this will resolve the issue after bumping the version of the Cypress gh action that we use to v6
- Maybe it won't, maybe it will : -)


Hoolean (approved):
Looks like we're well due on upgrade in either case :rocket:

I checked [the action's CHANGELOG](https://github.com/cypress-io/github-action/blob/master/CHANGELOG.md) and it looks like the main breaking changes were major Node version bumps; we're on 20 now, which is the current LTS, and so we should be fine here :)

ob6160:
Thanks for checking @Hoolean :)

PR #1728, opened 2024-10-06
Refactor Drawer component and remove framer-motion dependency

### LINKED ISSUE #1654: Refactor Drawer component and remove framer-motion dependency
**Is your feature request related to a problem? Please describe.**
Remove an additional dependency of framer-motion used only in the Drawer component.

**Describe the solution you'd like**
Remove the framer-motion dependency that is used only in this component for the animation and use css animation

**Additional context**
[See PR for discussion on this component](https://github.com/public-convenience-ltd/toiletmap/pull/1653)

Hoolean:
Closes #1654

## What does this change?

This commit replaces the `framer-motion` dependency with a custom CSS transition.

To keep the implementation simple, the `visibility` property is being toggled instead of `display`, as it still removes the element from the accessibility tree but has the benefit of being animatable. As the element has a fixed position, it should not have significantly worse performance.

(an aside: abandon hope all ye who  try to make it work with `display` again ☠️ this led me down a path of deeper and deeper problems, e.g. with elements being unable to start transitions while hidden, which needed more and more stateful React thrown against the wall to even come close to fixing)

## How was this tested?

I've only tested this locally so far; I'll leave this as a draft PR until I see if I can do something more robust :)

## Accessibility

   - Verified that `visibility` is semantically equivalent to `display` for this purpose
    - Tested on Vercel staging site
    - N/A, no design changes

## Documentation

  - Confirmed that no remaining references to framer
  - Already present


Hoolean:
(creating as a draft PR to track progress, will undraft as I address the testing, accessibility, and documentation checklists above, where appropriate)

ob6160:
thanks so much for this @Hoolean it'll be great to remove more deps over time as we simplify the client 👏

ob6160 (approved):
this is great, another dependency down 🫡

PR #1743, opened 2024-11-19
Add new Stack and Center layouts and List component to design system

rosejbon:
## What does this change?

This update introduces two core layout components: `Stack` and `Center`. These are designed to replace the current usage of `Box` and `Spacer` components, promoting a more consistent approach to content layout.

The `Stack` and `Center` layouts are inspired by the layouts of the same name from Andy Bell and Heydon Pickering's [Every Layout](https://every-layout.dev/layouts/).

I've updated the `about` page to use these new layouts to show how they would work.

For completions sake I added a `List` component to completely remove any use of Styled Components in that about page example.

## How was this tested?

- **End-to-end testing:** Verified using Cypress to ensure functionality remains intact.
- **Visual testing:** Used Chromatic to compare visual differences and confirm no unintended regressions.

## Accessibility

- I couldn't think of any accessibility impacts that this would have. 

## Documentation

-  All new components have been added to Storybook, complete with examples and usage instructions.


src/design-system/layout/Center/Center.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import centerProps from './Center.types';
+
+const Center: React.FC<centerProps> = ({
+  text = false,
+  gutter = true,
```
mxdvl, line 6: I wonder whether it's beneficial to use conditional props here over explicit ones. Are we losing much by making them required?

More importantly I think a small explanation of what they do would be beneficial to future travellers:

```suggestion
  /** used to center align text to the horizontal centre */
  text = false,
  /** include horizontal padding around the text */
  gutter = true,
```

src/design-system/components/List/ListGroup.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <>
+      {ordered ? (
+        <ol className="list">{children}</ol>
+      ) : (
+        <ul className="list">{children}</ul>
+      )}
+    </>
+  );
+};
```
mxdvl, line 18: Not entirely sure about the benefit of this component over simply exposing `ol` & `ul`. I fear this abstraction might bring about more confusion, for example:
- What's the difference between using the elements directly and the wrapper?
- What happens if some of my `children` are not `ListItem` components?
- Why is the default list unordered if I omit the attribute?

mxdvl (commented):
On the whole this is a good change and a good step towards removing styled components. Your thoroughness is commendable and visual testing very adequate.

The only reason for my lack of explicit approval is around `ListItem`. I do wonder if the direction to go is towards higher abstractions and custom components or whether we should make things more concrete and simply use good old HTML tags. I find the adage “solve for the problem you have today, not a future problem you imagine” a helpful framing, especially in projects with low change frequency. Equally, before promoting a pattern to a fully fledged component, I try to follow a “rule of three”, which entails waiting for three identical uses of a pattern/function/component before making it generic. This tends to prevent ending up with many parameters/props/configuration to adjust as the same thing is used for actually different uses. I hope some of this reflection is useful.

However, I'm only a casual contributor here and if others are happy for this to go in as-is, so am I. 😁

src/design-system/components/List/ListGroup.tsx
rosejbon: Hey Max, super appreciate your review. I get where you are coming from, and I did have similar reservations when making that component and those are fair points about confusion around the component. 

For context, these were the factors I considered: 

 - In our base css we remove any listing styling a browser would add
 - We need unstyled lists in serveral places, e.g the filter, the nav, etc.
 - We will likely need lists in our about pages and new blog

My options were:

1. Add a css class for lists that we just add to lists where we need styled lists
2. Reintroduce default list styling by updating the base css and temporarily fix the places where that breaks styling
3. Create a custom component for a list

I dismissed the first because it would break the patterns we've established in this design system. Whilst we are using classes, this is very much a "Styled Component" inspired design system where we don't envisage developers having to add classes to loads of things. 

The second, I also dismissed because I assumed it would be messy to refactor, but I might have dismissed that to easily.

The third I choose because it was something used both on the `login` and `about` pages, and could see being used in the new blog styling. 

That was a super long winded way of saying, you've convinced me to remove this component and instead go for the second option and use `ol` & `ul`. 😆

src/design-system/layout/Center/Center.tsx
rosejbon: Yeah, that’s a fair point and makes sense. I’ll update the component to use required props and include some documentation to make the behaviour clearer.

src/design-system/components/List/ListGroup.tsx
ob6160: I second the second option as it'll be super useful for when we get blog posts going, right now the blog setup uses just standard markdown (not MDX) so having styles that work well for instances of `ol` and `ul` instead of having the hassle of inserting the components into the MDX will be great.

rosejbon: I have messed with the timeline and like magic that List component never existed 🪄

I've returned us to default styling for lists, adding consistent margins using our spacing variables. Along the way, I addressed issues this caused in a few places—mainly parts I hadn’t yet refactored. So for now, I've updated the relevant styled components and plan to tidy this up properly when I come to them.

mxdvl (approved):
Love a good use of the web's primitives! 🚀

ob6160 (approved):
🚀

PR #1753, opened 2025-02-20
Add new header to design system

rosejbon:
## What does this change?

This branch has been sitting on my machine since Christmas 🙈 

This PR mainly adds the header to the design system, but in the process also refactors and updates a few things:

- **Visually Hidden Component:** Moved into the design system and refactored its usage throughout the codebase.
- **Expanded Center Layout Utility:**
   - Introduces an "article" prop for the blog and text based pages like the `about us` page.
   - Open to feedback on whether this layout is taking on too many responsibilities.
- **Mobile Navigation Cleanup:** Removed the footer from the mobile nav dropdown, as it felt unnecessary and semantically incorrect (a `<footer>` inside a `<header>`).
- **Header Adjustments:**
  - Removed the "Find a toilet" link, which only redirected to the homepage.
  - Adjusted the header layout on desktop to improve its visual balance.
  - New desktop layout: image
- **Accessibility Improvements:** Added a focus trap to the mobile dropdown menu to enhance keyboard navigation.

Let me know if you have any thoughts or suggestions! 🚀

## How was this tested?

- Tested on my local machine on Safari, FireFox, and Chrome
- Testing your change on a preview deployment and with Cypress tests

## Accessibility

- [x] Checked to see if navigable by keyboard (see changes to improve mobile keyboard navigation)

## Documentation


src/pages/about.page.tsx
```diff
@@ -105,37 +105,34 @@ const AboutPage = () => {
...
+                restaurants, hotels and pubs.
+              </li>
+              <li>
+                We try not to show toilets where you need to ask permission to
+                use it, such as those that are for customers-only. The exception
+                is platform-side toilets at stations. These are only accessible
+                to ticket-holders but we feel it is still useful information,
+                particularly when many trains don’t have toilets onboard.
+              </li>
+            </ul>
```
mxdvl, line 135: Is this diff intentional?

src/globalStyles.tsx
```diff
@@ -38,7 +38,7 @@ const globalStyles = (
         --text-4: clamp(2.07rem, calc(1.84rem + 1.17vw), 2.75rem);
         --text-5: clamp(2.49rem, calc(2.16rem + 1.64vw), 3.43rem);
 
-        --toilet-max-width: 845px;
+        --toilet-max-width: 1680px;
```
mxdvl, line 41: Why did we change this value?

src/design-system/components/Header/Header.tsx
mxdvl, line 1: Is this lifted from somewhere? Do we need to review its inner workings in detail?

src/pages/about.page.tsx
rosejbon: Yes, that is a good question, I'd forgotten why I'd fixed that, but it was causing a react hydration error because of the illegal use of nesting a <ul> in a <p> tags. 😅 Turns out I don't know the HTML spec as well as I thought I did, thought I could sweep my error under the cover rather than admitting it in a commit 🙈

src/globalStyles.tsx
rosejbon: It is a mildly arbitrary number I stole from the designers I work with. But because I wanted the `Center` component to be wider by default and have a max-width of `75ch`* for blogs and other text based content I edited this. I toyed with the idea of adding a global variable, something like `--toilet-width: 75ch`, maybe I should for clarity? 🤔 It rather hidden in that `Center` layout.

*This number does have rationality to it, taken from this article - https://piccalil.li/blog/line-length/

src/design-system/components/Header/Header.tsx
rosejbon: Yeah this is a rather unfortunate downside of moving the components to a new folder, hindsight being everything I probably wouldn't have done that now. It was taken from this deleted header - https://github.com/public-convenience-ltd/toiletmap/pull/1753/commits/74b2076de894f1c1d687d7c6eae982665971f91b#diff-08f1eeca099377b71b51c70a42d155ddf8d0d6152e3961b53402a482f70c280c

src/globalStyles.tsx
mxdvl: I like it, but maybe use `ch` units directly and add a CSS comment as to why with a link to that reference?

src/pages/about.page.tsx
mxdvl: Generally I'd suggest extracting these changes out to make it easier to review the PR, but not a blocker in this case.

mxdvl (approved):

PR #1758, opened 2025-03-24
bump next to 15.2.3

ob6160:


Hoolean (approved):
LGTM :shipit: Thanks for pouncing on this - agree with your assessment of the CVE.

Side note: out of curiosity, have you seen any detail on why Vercel was not affected?

ob6160:
> LGTM :shipit: Thanks for pouncing on this - agree with your assessment of the CVE.
> 
> Side note: out of curiosity, have you seen any detail on why Vercel was not affected?

@Hoolean here is the reason why 🫨 

> Yeah, our system sets this header before invoking the middleware and ignores user supplied headers. Hence no chance for an attacker to ever control it.
> Has nothing to do with the firewall. We just got lucky

https://x.com/cramforce/status/1903648110863343871

PR #1759, opened 2025-04-21
Feature: add new footer to design system

rosejbon:
## What does this change?

Quite a small bank holiday update, just adds the Footer to the new design system. Also removes sponsored by Vercel and replaces it with Support Us and a link to our Patreon account.

## How was this tested?

Cypress tests  👀

## Documentation


src/design-system/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,9 @@
+import { Meta } from '@storybook/react';
+import Footer from '.';
+
+export default {
+  name: 'Footer',
+  component: Footer,
+} as Meta;
```
mxdvl, line 7: **question:** did you mean `satisfies Meta`? Otherwise type with `as` casting opts out of the compiler checks so I would strongly recommend against it.

src/design-system/components/Footer/Footer.css
```diff
@@ -0,0 +1,33 @@
+.footer {
+    padding-block: var(--space-xs);
+}
```
mxdvl, line 3: **question:** why are we using a class name that's identical to the one element name? Why not use the `footer` selector directly?

rosejbon: Yeah, I get that it might seem redundant having a `.footer` class on a `<footer>` element, but for me it's about consistency and control. I like using class names even on semantic elements because it keeps things scalable and aligns with the BEM-ish structure I'm loosely following.

src/design-system/components/Footer/Footer.stories.tsx
rosejbon: Thanks for pointing that out! I updated the story to follow the [Storybook docs more closely](https://storybook.js.org/docs/api/csf#default-export) and explicitly typed `meta` as `Meta<typeof Footer>`. Appreciate the heads-up and it reminds me I need to fix up the other stories after this PR.

src/design-system/components/Footer/Footer.css
mxdvl: If we're creating a component library, when would we use a `footer` element that's not `Footer`? Control & scalability can come at the expense of simplicity and consistency…

But if that's something we've done elsewhere that's fine! 😊

Hoolean (approved):
Nice! LGTM 🎉

ob6160 (approved):

PR #1766, opened 2025-07-05
Data download page

ob6160:
## What does this change?

- We add a dedicated page with accompanying licence information so that users are able to download a copy of our complete dataset in either JSON or CSV format under CC BY 4.0
- Descriptive comment added to the PointInput type
- Consistent lower case names for page stories


rosejbon (approved):
Looks good to me 😄  :shipit:

